PR #3, opened 2011-12-25
Made inheritance trees more vanilla-like.

macalinao:


Raphfrk:
I'll leave this to @RoyAwesome , but I was thinking the same.

macalinao:
@RoyAwesome, I still added support for your sky entity or whatever by making living entity inherit from moving entity.

PR #10, opened 2012-01-23
Couple more events for VanillaAPI

Zidane:
Title says it all :P


zml2008:
You should add fields in the constructor for the event's included fields.

Zidane:
Alright, one moment

Zidane:
I hope that is better zml2008

PR #11, opened 2012-01-23
More event changes for VanillaAPI

Zidane:
Title says it all :P


macalinao:
Why did you remove the second asterisk from the javadocs...

Zidane:
Erm the headers are that way....

PR #13, opened 2012-01-28
Tree generation

windy1:
Made some normal trees, I attempted birch trees, but <code>VanillaBlocks.birchLog</code> and <code>VanillaBlocks.log</code> seem to be the same at the moment. I also had some encounters with some astray blocks, I don't think it has to do with me and probably the way chunks are loaded.

(The number of commits are due to formatting issues :/)

### DIFF
--- a/src/main/java/org/spout/vanilla/generator/normal/TreePopulator.java
+++ b/src/main/java/org/spout/vanilla/generator/normal/TreePopulator.java
@@ -33,124 +33,123 @@ import org.spout.api.geo.discrete.Point;
 import org.spout.vanilla.VanillaBlocks;
 
 public class TreePopulator implements Populator {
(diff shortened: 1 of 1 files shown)


Afforess:
Ugh, pulled prematurely. Please sign off on this, and squash the commits into one.

Afforess:
Also the reason why you could not see birch was because chunks are not sending block data yet.

windy1:
@Afforess Err, this is gonna sound kinda dumb but I'm having trouble with the interactive rebase gui. I make my changes, but then where do I go from there and exit the editor?

Like, I don't see any save and close option http://i.imgur.com/boY8I.png

windy1:
Figured it out, see the new pull

PR #16, opened 2012-01-28
Falling sand!

Darsstar:
No animation though =[


windy1:
We're already ahead of Minecraft PE :D

PR #30, opened 2012-02-04
Improved food, and you can now place redstone!

ZNickq:
Signed-off by ZNickq teapa011@gmail.com


src/main/java/org/spout/vanilla/item/GenericVanillaItemMaterial.java
```diff
@@ -45,6 +46,15 @@ public GenericVanillaItemMaterial(String name, int id, int data) {
 
 	@Override
 	public Material getBlock() {
+		if (this == VanillaMaterials.REDSTONE) {
```
zml2008, line 49: There's this thing called subclassing. Maybe you should try it.

Afforess, line 49: +1 Agree @zml2008

ZNickq, line 49: Well, i thought about it, but decided it would be a little overkill making new classes just for that :P But ok, will change!

Zidane, line 49: I agree with my teammates on this one as, despite you thinking its overkill, hard-coding usage of materials in an if structure such as that is never the answer and makes it very limited.

Once you finish that up, squash it into one (after reviewing it) please. Otherwise, I am afraid I can't pull this :/.

Zidane:
Closing this as @ZNickq forgot to :p

PR #32, opened 2012-02-12
Further command implementation/syntax

windy1:
Added a bit more syntax for the commands. Syntax tested and working for Spout.

Signed-off-by: WalkerCrouse wjcrouse913@aim.com

### DIFF
--- a/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
+++ b/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
@@ -25,38 +25,160 @@
  */
 package org.spout.vanilla.command;
 
+import org.spout.api.Spout;
 import org.spout.api.command.CommandContext;
 import org.spout.api.command.CommandSource;
 import org.spout.api.command.annotated.Command;
 import org.spout.api.command.annotated.CommandPermissions;
 import org.spout.api.exception.CommandException;
+import org.spout.api.player.Player;
 import org.spout.vanilla.VanillaPlugin;
 
 /**
  * Commands to emulate core Minecraft admin functions.
  */
 public class AdministrationCommands {
+	
 	public AdministrationCommands(VanillaPlugin plugin) {
+		
 	}
-
-	@Command(aliases = {"gamemode", "gm"}, usage = "[player 1/survival or 2/creative] or [1/survival or 2/creative] (if source is a player)", desc = "Change a player's gamemode", max = 2)
+	
+	@Command(aliases = {"gamemode", "gm"}, usage = "[player] <1|2> (1 = SURVIVAL, 2 = CREATIVE)", desc = "Change a player's gamemode", max = 2)
 	@CommandPermissions("vanilla.command.gamemode")
 	public void gamemode(CommandContext args, CommandSource source) throws CommandException {
-		source.sendMessage("Gamemode worked.");
-		//TODO implement gamemode.
+		// If source is player
+		if (args.length() == 1) {
+			if (source instanceof Player) {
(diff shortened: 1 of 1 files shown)


zml2008:
These command implementations would be terrible to maintain and use ints where string values should be used instead. Also, there are methods in CommandContext that handle int checking for you and for errors you throw a CommandException.

windy1:
@zml2008 Oh, I didn't know we wanted string values for modes and what not I can fix the other things though if you think it's worth it.

windy1:
Man, this is where I wish I had String switch statements now :(

windy1:
Okay, cleaned it up quite a bit

windy1:
Okay added String support

src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -25,38 +25,160 @@
...
 	@CommandPermissions("vanilla.command.xp")
 	public void xp(CommandContext args, CommandSource source) throws CommandException {
-		source.sendMessage("Xp worked.");
-		//TODO implement xp.
+		// If source is player
+		if (args.length() == 1) {
+			if (source instanceof Player) {
+				Player sender = (Player) source;
+				int amount = args.getInteger(0);
+				source.sendMessage("You have been given " + amount + " xp.");
```
Zidane, line 122: No reason to make a casted Player object and not even use it to send the message eh?

src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -25,38 +25,160 @@
...
+		
+		// If source is not player
+		if (args.length() == 2) {
+			Player player = Spout.getGame().getPlayer(args.getString(0), true);
+			if (player != null) {
+				int amount = args.getInteger(1);
+				player.sendMessage("You have been given " + amount + " xp.");
+				// TODO: Give player 'amount' of xp.
+			} else {
+				throw new CommandException(args.getString(0) + " is not online.");
```
Zidane, line 137: Anyone else hope to add xp to offline players? How about you @zml2008?

src/main/java/org/spout/vanilla/command/AdministrationCommands.java
windy1, line 122: That's for the to-come implementation of switching the gamemode which will require a player instance. (Doesn't really _need_ to be there atm)

Zidane:
Only reason I have not pulled this @Windwaker is that it is not neccessary to have some int checking method. Please observe these helper methods in CommandContext: http://jd.spout.org/latest/org/spout/api/command/CommandContext.html

windy1:
@NinjaZidane It's to see if it's a String or an int. I know I could just use args.getInteger(foo); but I'm checking so someone could use /gm creative or something similar. I still think it's justified.

Zidane:
True as well @WalkerCrouse, perhaps you could instead do a 

try{ getInteger(blah, blah)} catch

Use our stuff and not the scanner stuff :p

windy1:
But that's yucky :(, there should be an isInteger() in SpoutAPI then...

Zidane:
Yeah...now that I think about it.

Rebase it and I will pull it. We can modify it later if need be (besides the implementation).

windy1:
@NinjaZidane It's squashed... :p

Zidane:
No...I mean rebase it with the current head. I can't merge it automatically :p

windy1:
See https://github.com/SpoutDev/Vanilla/pull/33

PR #35, opened 2012-02-17
Implemented Weapon, RangedWeapon, made Food better!

ZNickq:
I thought for a while about the best way to implement ranged weapons, this seems like it! :)


Top-Cat:
What weapon isn't ranged and when would the two damage values be used from the ranged weapon?

ZNickq:
Any weapon except the bow, right now, isn't ranged...

-long answer that doesn't really make sense-

Top-Cat:
That was long winded and didn't really answer the question at all.

> "when would the two damage values be used from the ranged weapon?"

Are you saying that you want the damage value when hitting someone with a bow (not the arrow) to be modifiable?

ZNickq:
Ok, i'm saying that for a developer, it will be easier if he can easily get both the damage that a weapon does with a direct punch, as well as the damage when firing...

Top-Cat:
So yes to the question?

ZNickq:
Well, yes :D

src/main/java/org/spout/vanilla/material/generic/GenericRangedWeapon.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.spout.vanilla.material.RangedWeapon;
+
+public class GenericRangedWeapon extends GenericWeapon implements RangedWeapon {
+	
+	private int rangedDamage;
+	
+	public GenericRangedWeapon(String name, int id, int damage, int rangedDamage) {
+		super(name, id, damage);
+		this.rangedDamage=rangedDamage;
```
Top-Cat: spaces please, below too

ZNickq: Is it ok if i fix this in my next PR/in a separate commit in this one? Going to add some stuff to the identification handlers, and i'd hate to go through all the work of recreating the PR for some spaces :/

stephenwilliams: makes it a lot easier if you do each PR on a separate branch and not your master branch

Zidane: In IntelliJ, you can easily modify a prior commit (even if it is already pushed in a remote repo) as long as you use gitbash to force push to your personal fork.

It is how my commits come out flawlessly now in formatting....I can make sure things work well in my personal fork then push to SpoutDev when I am satisfied.

src/main/java/org/spout/vanilla/material/Weapon.java
```diff
@@ -27,4 +27,5 @@
 
 public interface Weapon extends Item {
 
+	public int getDamage();
```
Zidane, line 30: Having weapons automatically assume that they always do damage doesn't quite make sense to me...

What if some dev made a curing staff that is held differently (aka weapon) but when he cracks it over your head...it heals you :P.

That isn't damage...should probably change this to getAmount(); (same for rangedweapons....healing arrows xD).

src/main/java/org/spout/vanilla/material/generic/GenericRangedWeapon.java
stephenwilliams: > It is how my commits come out flawlessly now 

@alta189 goes to look for flaws

Zidane: Yeah...its possible as I might have some differences in my coding formatting than what you may have (as I went and tweaked it some).

stephenwilliams: I have changed it as well :p

Zidane: Probablyyyy need to post it on the forums (either of us) and we decide on one final template. THEN we get everyone to use it (I am sure ZNickq would love a template file to import into Netbeans/IntelliJ so he doesn't have to bother with formatting no more).

ZNickq: Actually, netbeans formats it right, i just forget to press format :P
@alta189 hmm, gonna try that branch thing, would be awesome if it works o_O

Anyway, my power/internet went down, so don't expect any changes for about 12 hours :P

src/main/java/org/spout/vanilla/material/Weapon.java
ZNickq, line 30: Well, if it's a "weapon" it kinda asumes it does damage, i thought about this :P
So, to be held differently, it should extend just GenericTool (not sure if that exists :P) :)

src/main/java/org/spout/vanilla/material/generic/GenericRangedWeapon.java
Zidane: That sucks :/

src/main/java/org/spout/vanilla/material/Weapon.java
Zidane, line 30: It does and I didn't quite think about that. I guess its better to say that GenericWeapon simply provides a forced reminder that this item will do damage and as such make sure you provide a way to get its' damage (and set other attributes potentially in the future).

src/main/java/org/spout/vanilla/material/generic/GenericRangedWeapon.java
ZNickq: My internet is back on :D
So, getting intelliJ just for that "gitbash" you talked about :)

PR #38, opened 2012-02-19
Ban / Whitelist feature

greatman:
Player is kicked when he is banned or not in the whitelist.

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapIdentificationMessageHandler.java
+++ b/src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapIdentificationMessageHandler.java
@@ -25,6 +25,10 @@
  */
 package org.spout.vanilla.protocol.bootstrap.handler;
 
+import java.util.Arrays;
+import java.util.Iterator;
+
+import org.spout.api.Server;
 import org.spout.api.event.Event;
 import org.spout.api.event.player.PlayerConnectEvent;
 import org.spout.api.player.Player;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapIdentificationMessageHandler.java
```diff
@@ -35,7 +39,39 @@
 public class BootstrapIdentificationMessageHandler extends MessageHandler<IdentificationMessage> {
 	@Override
 	public void handle(Session session, Player player, IdentificationMessage message) {
+		
+		//We check if the player is banned
+		String playerName = message.getName();
+		
+		Server s = (Server) session.getGame();	
+		
+		Iterator<Player> bannedPlayerList = s.getBannedPlayers().iterator();
```
Afforess: This is a terribly slow way to find out if a player is banned or whitelisted. What if there are 10000 players on the whitelist? The API needs to be changed to use Set, backed by a HashSet. Then you can do a contains call and have it be O(1) vs the current O(n) lookup time.

ZNickq: Agree...i'd do it, but that'll put multiple commits in my PR right now, and i suck at squashing them :P

stephenwilliams: Again @ZNickq if you do all of your PRs in separate branches you would be ok

ZNickq: Oh, that's an idea, netbeans lets me do that easily! :D
Ok, will do in the evening :)

windy1: git + netbeans = :(

windy1:
Needs to be squashed and signed.

windy1:
Also needs to be updated to current head which includes a new configuration.

Salynne:
Closing these, open a new one if you've updated them.

Zidane:
This will be closed so you can fix it and come back with a better pull

Thank you for your contribution regardless though...much appreciated :).

PR #40, opened 2012-02-21
Vanilla now checks for the right protocol id, and only allows connecting if that matches!

ZNickq:


Zidane:
I might have pulled this BUT we have been discussing allowing clients to connect on differening Vanilla protocol versions.

aka someone with Vanilla client 1.2 could play at the same time on the same server (running Spout 1.2 RB with Vanilla 1.2 RB) with a few people running 1.` or even 1.0 client.

ZNickq:
I thought about that, but there are protocol differences which make that hard (1.2 has different chunk protocol, for example) and i'm not sure it could work without keeping the old code...

Zidane:
Well we are gonna try ZNickq and see...unsure yet which way we are gonna head to.

ZNickq:
well, this can just sit here in the meantime :)

Afforess:
@NinjaZidane - I think I agree with @ZNickq on this one.

Zidane:
With all the changes occurring to Minecraft 1.2....I am starting to agree as well. It might just be too difficult to put in backwards compatibility :/

So, in light of this I am gonna go ahead and merge this.

PR #41, opened 2012-02-21
Some changes to the networking part, changed PlayNote to BlockAction and updated EntityAction a little!

ZNickq:
For anyone wondering, the notchian client uses a "Play Note" packet, with custom data instead of the instrument and pitch, for piston movement! So it doesn't really make sense calling it Play Note, could confuse people :P

P.S: @alta189 i love branches! :D


src/main/java/org/spout/vanilla/protocol/msg/BlockActionMessage.java
```diff
@@ -50,16 +52,16 @@ public int getZ() {
...
-	public int getPitch() {
-		return pitch;
+	public int getSecondByte() {
+		return secondByte;
 	}
 
 	@Override
 	public String toString() {
-		return "PlayNoteMessage{x=" + x + ",y=" + y + ",z=" + z + ",instrument=" + instrument + ",pitch=" + pitch + "}";
+		return "PlayNoteMessage{x=" + x + ",y=" + y + ",z=" + z + ",firstByte=" + firstByte + ",secondByte=" + secondByte + "}";
```
Zidane, line 65: Me thinks you forgot BlockAction here...

PR #46, opened 2012-02-24
Repost of PR 34

Darsstar:
https://github.com/SpoutDev/Vanilla/pull/34

Had a mindfart or something...


Zidane:
The formatting seems a little screwy on this PR (some classes being wiped and redone is an indicator).

Also, how about GenericFullContainerItem as a name instead of what you have? As it is a block in the world, item when a player has it and sometimes an entity (if it moves).

PR #49, opened 2012-02-29
Used SpoutEntity.is(....) instead of SpoutEntity.getController() instanceof

Darsstar:


Zidane:
I talked with @Afforess and we wanted to remove the .is method. Instanceof check is actually faster.

Thanks for the pull request though :).

PR #50, opened 2012-02-29
Updated the gamemode command and the IdentificationMessage now sends the...

Darsstar:
... gamemode based on the players' Controller instance.

new: 0 is survival and 1 is creative
old: 1 is survival and 2 is creative

If the player's controller is and instance of SurvivalPlayer you will spawn in survival mode, creative otherwise.


Zidane:
This pull is fine though :).

PR #51, opened 2012-02-29
Added kick command.

spacechase0:
I assume this is the right way to do this (both the pull request and the command)? :P

### DIFF
--- a/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
+++ b/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
@@ -111,6 +111,19 @@ public class AdministrationCommands {
 			source.sendMessage(player.getName() + "'s game mode has been changed to " + message);
 		}
 	}
+	
+	@Command( aliases = "kick", usage = "[player]", desc = "Kick a player from the server", max = 1 )
+	@CommandPermissions( "vanilla.command.kick" )
+	public void kick( CommandContext args, CommandSource source ) throws CommandException {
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -111,6 +111,19 @@ public void gamemode(CommandContext args, CommandSource source) throws CommandEx
 			source.sendMessage(player.getName() + "'s game mode has been changed to " + message);
 		}
 	}
+	
+	@Command( aliases = "kick", usage = "[player]", desc = "Kick a player from the server", max = 1 )
```
windy1, line 115: We don't use spaces in between parentheses like that it should be, <code>@Command(aliases = "kick" ...)</code>
Same goes for the rest of the pull (for the future)

windy1:
Kick is a Spout command and has already been implemented.

Zidane:
@WalkerCrouse speaks the truth.

spacechase0:
Aww. Guess I'll look for something else. :P Trying to find something small, since I don't know much of the codebase.

Zidane:
Don't feel put off @spacechase0, this bit of code just doesn't fit in Vanilla. If you get another idea for an improvement...don't hesitate to put up another PR!

Darsstar:
And just like you can set the maximum allowed parameters with `max` you can set the minimum required parameters with `min`.
Then you don't need that `if (args.length() == 1)`

PR #52, opened 2012-02-29
Added mob and vehicle spawn protocol /Bergerkiller

bergerkiller:
Adds the base protocol classes for all mob and vehicle types. The mob types with additional meta data during creation (parameters) reside in separate classes for proper parsing, the others reside in 'GeneralLivingEntityProtocol'

### DIFF
--- a/src/main/java/org/spout/vanilla/entity/protocols/living/GeneralLivingEntityProtocol.java
+++ b/src/main/java/org/spout/vanilla/entity/protocols/living/GeneralLivingEntityProtocol.java
@@ -1,0 +1,97 @@
+/*
+ * This file is part of Vanilla.
+ *
+ * Vanilla is licensed under the SpoutDev License Version 1.
+ *
+ * Vanilla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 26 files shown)


src/main/java/org/spout/vanilla/entity/protocols/living/GeneralLivingEntityProtocol.java
```diff
@@ -0,0 +1,97 @@
...
+		} else if (c instanceof Wolf) {
+			type = 95;
+		} else if (c instanceof Mooshroom) {
+			type = 96;
+		} else if (c instanceof SnowGolem) {
+			type = 97;
+		} else if (c instanceof Ocelot) {
+			type = 98;
+		} else if (c instanceof Villager) {
+			type = 120;
```
zml2008, line 89: This is horrible

bergerkiller:
@zml2008 Can't see why, as I discussed doing that on IRC. Or do you want separate protocol class files for every single mob?

src/main/java/org/spout/vanilla/entity/protocols/living/GeneralLivingEntityProtocol.java
RoyAwesome, line 89: For reference, these numbers are stored in the entities datatable

windy1:
Could you squash and update to current HEAD please? :)

ZNickq:
Closed on request!

PR #56, opened 2012-03-02
Send weather updates to clients.

spacechase0:
Should work for natural weather changes, too, but I was testing with my /weather command. :P


RoyAwesome:
Thank you kindly sir.

PR #61, opened 2012-03-02
Minor time fix.

spacechase0:
Before, the time would have getting bigger, even past 24000, since it was checking `tickcount` instead of `time`. After fixing that, `tickcount` wasn't being used for anything.


windy1:
Herp

PR #62, opened 2012-03-02
Changed the owner of the commands to be the plugin itself

Nefaro:
Signed-off-by: Nefaro xsvenson@gmail.com


windy1:
Compiled and tested?

PR #65, opened 2012-03-03
Added sheep meta-data.

spacechase0:
Colored Sheep

### DIFF
--- a/src/main/java/org/spout/vanilla/VanillaPlugin.java
+++ b/src/main/java/org/spout/vanilla/VanillaPlugin.java
@@ -43,6 +43,8 @@ import org.spout.api.math.Vector3;
 import org.spout.api.plugin.CommonPlugin;
 import org.spout.api.protocol.Protocol;
 import org.spout.vanilla.command.AdministrationCommands;
+import org.spout.vanilla.command.TestCommands;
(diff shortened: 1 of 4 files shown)


windy1:
We don't want multicolored sheep to spawn by default, only the Vanilla ones. Could you make the color be passed as a parameter through the constructor? Also, your formatting is off.

``` java
public void foo() 
{
    bar.foo();
}
```

Should be 

``` java
public void foo() {
    bar.foo();
}
```

Could you fix these please?

spacechase0:
I keep getting a null pointer exception when trying to do it in the constructor, I guess since `parent` isn't set yet.

``` java
parent.setData("SheepSheared", false);
```

Would a setColor/setSheared method be ok, instead? I'll make it default to white in `onAttached()`.

windy1:
Don't actually set it in the constructor, set a variable for later use

``` java
private final int color;

public Sheep(int color) {
    this.color = color;
}

@Override
public void onAttached() {
    // blah blah blah
    parent.setData("SheepColor", color));
}
```

Would be nice if you could overload the constructor as well so no args can be given for a white sheep.

spacechase0:
Oh, ok.

spacechase0:
Done.

Hope I didn't miss any more formatting stuff. :P

PR #68, opened 2012-03-04
Added more controllers and added them to the Entity enum.

Darsstar:
Added more controllers and added them to the Entity enum.
Added Tameable and Tamed interfaces.
Added Dog and Cat entities.
Refactored /give.

Also did https://github.com/SpoutDev/Vanilla/pull/66

### DIFF
--- a/src/main/java/org/spout/vanilla/VanillaPlugin.java
+++ b/src/main/java/org/spout/vanilla/VanillaPlugin.java
@@ -58,7 +58,7 @@ import org.spout.vanilla.protocol.VanillaProtocol;
 import org.spout.vanilla.protocol.bootstrap.VanillaBootstrapProtocol;
 
 public class VanillaPlugin extends CommonPlugin {
-	
+
 	public static final GameMode defaultGamemode = GameMode.SURVIVAL;
 	public static final int minecraftProtocolId = 28;
 	public static int vanillaProtocolId;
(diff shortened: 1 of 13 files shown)


src/main/java/org/spout/vanilla/VanillaPlugin.java
```diff
@@ -114,7 +114,7 @@ public void onEnable() {
 		config.load();
 
 		//Register commands
-		CommandRegistrationsFactory<Class<?>> commandRegFactory = new AnnotatedCommandRegistrationFactory(new SimpleInjector(), new SimpleAnnotatedCommandExecutorFactory());
+		CommandRegistrationsFactory<Class<?>> commandRegFactory = new AnnotatedCommandRegistrationFactory(new SimpleInjector(this), new SimpleAnnotatedCommandExecutorFactory());
```
windy1, line 117: This isn't really necessary...

Darsstar, line 117: True, but to me it was either make all command methods static so the SimpleInjector instance could be removed (since that object isn't really necessary either), or pass in the plugin which was used by both command classes anyway.

windy1, line 117: Fair enough, all in all it doesn't really matter.

PR #70, opened 2012-03-04
Added spawn eggs.

spacechase0:


src/main/java/org/spout/vanilla/material/item/SpawnEgg.java
```diff
@@ -0,0 +1,76 @@
...
+
+public class SpawnEgg extends GenericItem {
+
+	public SpawnEgg(String name, int id, int data) {
+		super(name, id, data);
+	}
+	
+	@Override
+	public void onInteract(Entity entity, Point position, Action type, BlockFace clickedFace) {
+		Class<? extends VanillaEntity> controller = org.spout.vanilla.entity.Entity.getByID((int) getData()).getController();
```
zml2008: I'd move this into the constructor, since there is one instance of SpawnEgg per damage value. Also, you only updated the sheep spawn egg. Was that intentional?

spacechase0: Good idea.

I didn't think the others would work, for some reason. Just tried it, and they do. :P I'll add those.

src/main/java/org/spout/vanilla/material/item/SpawnEgg.java
```diff
@@ -0,0 +1,80 @@
...
+		super(name, id, data);
+		
+		Class<? extends VanillaEntity> controller = org.spout.vanilla.entity.Entity.getByID((int) getData()).getController();
+		
+		Constructor<?> tmp = null;
+		Constructor<?>[] constructors = controller.getConstructors();
+		for (Constructor<?> constructor : constructors) {
+			if (constructor.getParameterTypes().length == 0) {
+				tmp = constructor;
+				break;
```
Darsstar: You could return instead of breaking.

Top-Cat: But then chosen wouldn't equal tmp ;)

Darsstar: Well, I figure he made `tmp` since he made `chosen` final. If he returns he can replace `chosen = tmp` by `chosen = null` and `tmp = constructor` by `chosen = constructor`, right?

spacechase0: I originally added `tmp` because when I just had `chosen = constructor`, it was complaining about `chosen` not getting initialized sometimes.

Do I really need to make another commit over such a minor detail though? :P

Darsstar: Ask @NinjaZidane, I just wanted to let you know. Oh, sorry for the mail @NinjaZidane :p

Zidane:
Heh...>.>

spacechase0:
Is that a yes or no? :P

Zidane:
@spacechase0 I am sure you can do it really quick...you will be fine haha.

spacechase0:
Fixed, for you picky people. :P

PR #78, opened 2012-03-07
Added sheep shearing.

spacechase0:
The sheep don't update client-side as sheared, and the item data also isn't sent correctly to the client.

### DIFF
--- a/src/main/java/org/spout/vanilla/VanillaMaterials.java
+++ b/src/main/java/org/spout/vanilla/VanillaMaterials.java
@@ -67,6 +67,7 @@ import org.spout.vanilla.material.item.Minecart;
 import org.spout.vanilla.material.item.PoweredMinecart;
 import org.spout.vanilla.material.item.RedstoneTorch;
 import org.spout.vanilla.material.item.RedstoneWire;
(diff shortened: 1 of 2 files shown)


windy1:
Looks good to me @NinjaZidane :)

I'm guessing the updating client side is Spout not updating the entity data correctly?

Zidane:
Is it just me or is @spacechase0 's obession with sheep...spooky.

Darsstar:
Not just you, my theory is that he will revert movement next to sent the sheeps to his alien overlords. I guess that technically makes me think his obsession with sheep is alien.

spacechase0:
Sheep had several things needed doing. :P They have a lot more that need doing than, for example, pigs.

> I'm guessing the updating client side is Spout not updating the entity data correctly?
> Yes, at least as far as I know. It only sends those things in `getSpawnMessage`, and `getUpdateMessage` only sends `EntityTeleportMessage`.

One way I guess would be for `getUpdateMessage` to return a List (and change the name to plural :P), so that I could add a `EntityMetadataMessage`. Then `getSpawnParameters` could be renamed to `getMetadataParameters`, and that could be used for `getSpawnMessage`, too.

PR #85, opened 2012-03-08
Added mob drops.

spacechase0:

### DIFF
--- a/src/main/java/org/spout/vanilla/entity/LivingEntity.java
+++ b/src/main/java/org/spout/vanilla/entity/LivingEntity.java
@@ -25,6 +25,40 @@
  */
 package org.spout.vanilla.entity;
 
+import java.util.Random;
+import java.util.Set;
+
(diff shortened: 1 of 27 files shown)


src/main/java/org/spout/vanilla/entity/living/passive/Villager.java
```diff
@@ -25,15 +25,28 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import org.spout.api.inventory.ItemStack;
 import org.spout.vanilla.entity.Entity;
 import org.spout.vanilla.entity.living.Creature;
 import org.spout.vanilla.entity.living.creature.Passive;
+import org.spout.vanilla.VanillaMaterials;
 
-public class Villager extends Creature implements Passive { // (Squidward)
```
windy1: :(

RoyAwesome: hrrnngghhhh

PR #87, opened 2012-03-08
Added a NetBeans license hint to the pom

Darsstar:
see http://wiki.netbeans.org/MavenBestPractices#License_Header

Is this this @Wulfspider approved?


ZNickq:
The formatting is wrong ;)

Darsstar:
Maybe I should have waited until class was over and pay more atention to both things. If this train would let my laptop onto the network I could fix it right now...

Darsstar:
It is now indented with tabs instead of spaces like the the poms of Spout and SpoutAPI

lukespragg:
This isn't really necessary, being as we may be pushing project files for each. This entry in the pom would only apply to anyone who actually manually made that license in their IDE, so a full project file would be better.

PR #95, opened 2012-03-18
Compatibility update for SpoutAPI block data change PR

bergerkiller:

This pull restores compatibilty with the pull found here:
https://github.com/SpoutDev/Spout/pull/1761

Changes include:
- Proper constants for Wool, Dye, Potion, Slab, DoubleSlab, and others with sub-data
- Slab can now be stacked and form one DoubleSlab
- VanillaMaterials constants updated to contain the root types only
- Functions using 'getTypeId()' or others changed to use getmaterial().getId()
- Functions changed to use Blocks instead of recursive xyz changes (tree decorator)
- Lots of general error fixes other than the above


src/main/java/org/spout/vanilla/VanillaMaterials.java
```diff
@@ -250,8 +209,8 @@
 	public static final ItemMaterial RED_APPLE = new GenericFood("Apple", 260, 4, FoodEffectType.HUNGER);
 	public static final ItemMaterial BOW = new GenericRangedWeapon("Bow", 261, 1, 9, (short) 385);
 	public static final ItemMaterial ARROW = new GenericItem("Arrow", 262);
-	public static final ItemMaterial COAL = new GenericItem("Coal", 263, 0);
-	public static final ItemMaterial CHARCOAL = new GenericItem("Charcoal", 263, 1);
+
```
windy1: What's up with the random new lines in some place? (Formatting complex) :p

bergerkiller: Yeah you are right, I shall remove some of those. Had to split the materials up sometimes to find certain 'TODO' area's back.

PR #97, opened 2012-03-20
Fix circular reference issues

bergerkiller:

There were some circular reference (actually, constructor called in constructor) in the Dye and other classes. For simplicity; all sub-mats now have on separate Main class extending it, to prevent this. I noticed you used the static 'PARENT' before, but couldn't figure out why that would be useful.

You can always turn the separate 'constants' into an ' = this' assignment if that is better.


Top-Cat:
Find out more about this PR here: https://github.com/SpoutDev/Vanilla/pull/97

windy1:
@Top-Cat wut? ... You just linked to this PR :p

PR #99, opened 2012-03-22
Swamp and Tree-vines

dpendic:

Added the Swamp biome and the vine covered trees that go into them


dpendic:
Due to problems with data values, the swamp just looks a softened plain biome with VERY distorted trees that have oddly floating vines

DDoS:
I too have some problems with metadata, with huge mushroom for the mushroom decorator. Good to know it's a Spout problem, and not my code :P.

dpendic:
Will include in Biome Selection PR

PR #100, opened 2012-03-23
Biome Selector overhaul and Swamp

dpendic:

Added BiomeSelection by "temperature" and "moisture", so that we don't have a desert and a frozen waist-land next to each other.

Also added Swamp biome. Trees look funky because of a problem with metadata on the Vines.

### DIFF
--- a/src/main/java/org/spout/vanilla/generator/normal/decorator/TreeDecorator.java
+++ b/src/main/java/org/spout/vanilla/generator/normal/decorator/TreeDecorator.java
@@ -73,28 +73,14 @@ public class TreeDecorator implements BiomeDecorator {
 		for (int k = 1; k <= oneWidth; k++) {
 			for (int i = -1; i <= 1; i++) {
 				for (int j = -1; j <= 1; j++) {
-<<<<<<< HEAD
-					if(w.getBlockMaterial(cx + i, cy + height - k, cz + j).equals(VanillaMaterials.AIR)) {
-						w.setBlockMaterial(cx + i, cy + height - k, cz + j, VanillaMaterials.LEAVES, c.getWorld());
-					}
-=======
 					b.clone().move(i, -k, j).setMaterial(VanillaMaterials.LEAVES);
->>>>>>> materials
 				}
 			}
 		}
 		for (int k = oneWidth + 1; k <= oneWidth + twoWidth; k++) {
 			for (int i = -2; i <= 2; i++) {
 				for (int j = -2; j <= 2; j++) {
-<<<<<<< HEAD
-					if(w.getBlockMaterial(cx + i, cy + height - k, cz + j).equals(VanillaMaterials.AIR)) {
-						if(!(j == -2 && i == -2) && !(j == 2 && i == -2) && !(j == -2 && i == 2) && !(j == 2 && i == 2)) {
-							w.setBlockMaterial(cx + i, cy + height - k, cz + j, VanillaMaterials.LEAVES, c.getWorld());
-						}
-					}
-=======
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/generator/normal/biome/UndefinedBiome.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.spout.vanilla.generator.normal.biome;
+
+import org.spout.api.util.cuboid.CuboidShortBuffer;
+import org.spout.vanilla.VanillaMaterials;
+import org.spout.vanilla.generator.VanillaBiomeType;
+
+public class UndefinedBiome extends VanillaBiomeType {
+	public UndefinedBiome() {
+		super(99);
```
zml2008: This has to be defined in the client or else clients will crash when receiving chunks containing this biome

src/main/java/org/spout/vanilla/generator/NoiseSelector.java
```diff
@@ -36,33 +36,39 @@
...
-		//Pick a biome at 256 height for both x and z
-		return (int) (noise.GetValue(x / 256.0 + 0.05, y + 0.05, z / 256.0 + 0.05) * 64);
+		tempBase.setSeed((int) seed*3);
+		moistBase.setSeed((int) seed^2); //math is to differentiate the seeds
+		
+		int biome = 99;
+		VanillaBiomeType biomes[] = {VanillaBiomes.DESERT, VanillaBiomes.JUNGLE, VanillaBiomes.MOUNTAIN, VanillaBiomes.MUSHROOM, VanillaBiomes.OCEAN, VanillaBiomes.PLAIN, VanillaBiomes.SWAMP, VanillaBiomes.TAIGA, VanillaBiomes.TUNDRA};;
+		for(VanillaBiomeType b : biomes) {
+			if(b.isValidPlacement(tempBase.GetValue(x, y, z)*100.0, moistBase.GetValue(x, y, z)*100.0)) {
+				biome = b.getBiomeId();
```
zml2008: Biome IDs are only for the vanilla protocol. BiomeSelectors do not know which biome they are actually selecting and cannot without a bit of refactoring.

src/main/java/org/spout/vanilla/generator/VanillaBiomeType.java
```diff
@@ -39,4 +43,17 @@ protected VanillaBiomeType(int biomeId, BiomeDecorator... decorators) {
 	public int getBiomeId() {
 		return biomeId;
 	}
+	
+	public boolean isValidPlacement(double temp, double moist)
+	{
```
zml2008: Braces are on the same line as the method declaration, not the line after.

src/main/java/org/spout/vanilla/generator/normal/decorator/TreeDecorator.java
```diff
@@ -56,6 +62,11 @@ public void populate(Chunk chunk, Random random) {
 	}
 
 	private void generateSmallTree(Chunk c, Random ra, int cx, int cy, int cz) {
+		BiomeGenerator bg = null;
+		if(c.getWorld().getGenerator() instanceof BiomeGenerator) {
+			bg = (BiomeGenerator) c.getWorld().getGenerator();
+		}
```
zml2008: Better would be subclassing TreeDecorator with each one having a method that takes coordinates, a chunk, and a random to generate each tree type, then a generic ChoiceDecorator which would choose a tree type from those available in the constructor for the biome.

PR #101, opened 2012-03-23
Biome Selection Overhaul

dpendic:

Biomes are now selected by a temperature and moisture value, which can be defined by a biome.

### DIFF
--- a/src/main/java/org/spout/vanilla/generator/NoiseSelector.java
+++ b/src/main/java/org/spout/vanilla/generator/NoiseSelector.java
@@ -25,10 +25,20 @@
  */
 package org.spout.vanilla.generator;
 
-import net.royawesome.jlibnoise.module.modifier.Turbulence;
-import net.royawesome.jlibnoise.module.source.Voronoi;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
 
+import net.royawesome.jlibnoise.NoiseQuality;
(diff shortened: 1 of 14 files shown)


src/main/java/org/spout/vanilla/generator/NoiseSelector.java
```diff
@@ -49,20 +55,38 @@
...
 	public int pickBiome(int x, int y, int z, long seed) {
-		base.setSeed((int) seed);
-		noise.setSeed((int) seed);
-		//Pick a biome at 256 height for both x and z
-		return (int) (noise.GetValue(x / 256.0 + 0.05, y + 0.05, z / 256.0 + 0.05) * 64);
+		temp.setSeed((int) seed);
+		humid.setSeed((int) seed);
+		double t = (temp.GetValue(x, y, z) * 50.0) + 50.0;
+		double h = (humid.GetValue(x, y, z) * 50.0) + 50.0;
+		VanillaBiomeType biome = VanillaBiomes.DESERT;
```
Zidane: I think this was accidentally left here

dpendic: Im meaning to change that to ocean after it's been finished. It's just a fallback incase no biome can be decided.

Zidane: Ah I see whats going on now...carry on :).

PR #102, opened 2012-03-23
VanillaConfiguration and OpConfig now set the value of nodes to the valu...

Darsstar:
...e in config.yml and ops.yml respectively if it exists.

Signed-off-by:Darsstar darsstar@gmail.com

See https://github.com/SpoutDev/SpoutAPI/pull/142 and https://github.com/SpoutDev/Spout/pull/1801

### DIFF
--- a/src/main/java/org/spout/vanilla/configuration/OpConfig.java
+++ b/src/main/java/org/spout/vanilla/configuration/OpConfig.java
@@ -43,6 +43,10 @@ public class OpConfig extends Configuration {
 	@Override
 	public void load() {
 		super.load();
+		Object value = getValue(node.getPath());
+		if (value != null) {
+			node.setValue(value);
+		}
 		this.addNode(node);
 		this.save();
 	}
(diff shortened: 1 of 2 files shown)


windy1:
This is unecassary. If the value is null it will fall back on the default value of the node which is defined in the constructor and through setDefaultValue()

windy1:
Oh I see... If this is having a problem it needs to be done in SpoutAPI

Zidane:
I will go ahead and pull this @W1ndwaker ...either of you are free to fix it in SpoutAPI and remove these two changes (referencing Spout as well).

windy1:
The problem is that I was loading the node value before the config value and returning that. It needs to be the other way around, I'll fix it next chance I have.

Darsstar:
Also the this pull (and the Spout one) rely on getNode() returning null not a Configuration node with the value null, which I changed in the SpoutAPI pull.
@W1ndwaker got any comments on the SpoutAPI pull?

PR #106, opened 2012-03-25
Added minecart physics

bergerkiller:

Part is still under development for a bit, but the main core is finished. It adds several functions extending classes can override to handle velocity modifiers, and others. The only thing that needs to be done is having 'y-snap' (snap to correct Y position) and some debugging.

### DIFF
--- a/src/main/java/org/spout/vanilla/controller/object/vehicle/Minecart.java
+++ b/src/main/java/org/spout/vanilla/controller/object/vehicle/Minecart.java
@@ -25,14 +25,273 @@
  */
 package org.spout.vanilla.controller.object.vehicle;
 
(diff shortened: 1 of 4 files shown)


Top-Cat:
There's a few spaces hanging around (need to be tabs) and not signed

bergerkiller:
Sec I'll fix it really quick...

EDIT

This should fix it...though I fail to sign of this PR itself.

Top-Cat:
Signed, but still got spaces

bergerkiller:
But where?! I did a global '    ' search and found none :/

Top-Cat:
http://i.imgur.com/xfZgs.png

bergerkiller:
Ok hopefully it is now fixed? xd

PR #108, opened 2012-03-25
Mushroom and huge mushroom generation code

DDoS:
I'm still working on the populate method. I've made the generation code flexible, so size parameters can be modified and the mushrooms will keep correct shape and metadata and new mushroom types can easily be added. Since metadata doesn't yet work on Spout, I test it it on CraftBukkit. I'm using world instead of chunk parameters for generation since a mushroom on the edge of a chunk can have blocks outside of it, which currently aren't being placed properly. This code also works for generating huge mushrooms by bonemeal using on small mushrooms.


DDoS:
Forgot to sign this off..

ZNickq:
You can sign it off w/o recreating the PR ;P

DDoS:
Too late...

PR #113, opened 2012-03-29
Fixed trees (again)

dpendic:

Makes trees the same as Vanilla. This was reverted when the materials branch was merged in.


bergerkiller:
Mhh why clone and move the same block twice? Might be easier/more efficient to store the cloned and moved block, use getMaterial() on it twice checking if it is air/leaves, and then setting the material. Just my opinion though :)

dpendic:
@bergerkiller I thought of doing that originally, but for some reason decided not to. Just amended.

PR #117, opened 2012-04-04
Basic Dropimplementation + Item not flying around like a bee and actually works.

steffengy:
Implemented basic drops of the entity, while they die.

### DIFF
--- a/src/main/java/org/spout/vanilla/controller/VanillaController.java
+++ b/src/main/java/org/spout/vanilla/controller/VanillaController.java
@@ -34,7 +34,9 @@ import org.spout.api.collision.CollisionModel;
 import org.spout.api.collision.CollisionStrategy;
 import org.spout.api.entity.Entity;
 import org.spout.api.entity.action.ActionController;
+import org.spout.api.entity.PlayerController;
 import org.spout.api.geo.cuboid.Block;
+import org.spout.api.geo.discrete.Point;
 import org.spout.api.inventory.ItemStack;
 import org.spout.api.math.MathHelper;
 import org.spout.api.math.Quaternion;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/controller/VanillaController.java
```diff
@@ -75,12 +76,24 @@ public void onAttached() {
 		getParent().setCollision(new CollisionModel(area));
 		getParent().setData(VanillaControllerTypes.KEY, getType().getID());
 	}
+	
+	/**
+	 * Called like onDeath, but before the entity gets actually destroyed
+	 */
+	public void preDeath() {
```
windy1: private void**

Zidane:
Lets agree on a onDeath instead of preDeath (preDeath is kind of a false impression).

steffengy:
As I remember there were two problems with using onDeath:
- The Entity didn't provide any Position
- was also called on Disconnect of Player

But If you're sure that using it, should be perfect Let me know.

ZNickq:
The animation for dropped items IS done by the client, i implemented it myself, but when you initially destroy a block, the item jumps a little, check normal mc :P

steffengy:
But your implementation forced the item to fly around in the air (like a crazy drunk bee) and as long as you don't implement it correctly you shouldn't break items completly ^^

ZNickq:
No, i implemented the animation sending, the flight code was done by someone else to test entity movement, among others...

steffengy:
Ah now i understand ^^

Zidane:
Steffengy it should be fine. The issue with disconnecting players calling onDeath can be dealt with. Quite simply by seeing if the player no longer has an assigned network synchronizer should do so. The player's entity is removed before the Pre-Snapshot stage of the tick (if the player logged out) so it works out.

As for your initial point...can you elaborate?

steffengy:
Will try to

windy1:
Has this been resolved? If not, you need to update to the current HEAD ;)

steffengy:
Already did once - this request was simply ignored then ^^

stephenwilliams:
@steffengy it needs to be rebased to the current HEAD before @NinjaZidane pulls it

PR #120, opened 2012-04-14
Refactored WindowClickMessageHandler.java

VictorD:
Hi! I've refactored WindowClickMessageHandler with more comments and extracted the logic into methods to make it easier to follow.

Any questions or comments you can reach me on IRC, (nick: Lightweight)


ZNickq:
Looks nice (i wrote the original version of that class :P) , but have you properly tested it?

windy1:
Having too many instance variables is bad, mkay. Could you try and use as many local variables as possible, several of those we don't need to store.

PR #121, opened 2012-04-14
Add methods for setting and getting hunger related vars.

MWisBest:
I thought this would be useful for plugin developers and possibly other things down the line. :)


MWisBest:
Do I have to update to the latest HEAD?

PR #124, opened 2012-04-14
Refactoring v3, WindowClickMessageHandler.java , now with less lines!

VictorD:
Evenin' , this PR aims to make it easier to maintain the WindowClickMessageHandler code by adding a few comments and extracting the logic into methods where possible.


ZNickq:
I'll have to manually test this out, i can see some cases where it could fail, but no way to tell for sure :) So it'll have to wait for a couple of days :<

Zidane:
@ZNickq 

Status update?

PR #129, opened 2012-04-15
Fix game mode message on connect. Fix game mode command. 

MWisBest:
On connect, a message was sent to the player saying their game mode was changed. This fixes that.

Also, the game mode command did not work with simply "/gamemode creative". This fixes that as well.

### DIFF
--- a/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
+++ b/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
@@ -255,7 +255,7 @@ public class AdministrationCommands {
 				throw new CommandException(args.getString(0) + " is not online.");
 			}
 		} else {
-			if (!(source instanceof VanillaPlayer)) {
+			if (!(source instanceof Player)) {
 				throw new CommandException("You must be a player to toggle your game mode.");
 			}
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -255,7 +255,7 @@ public void gamemode(CommandContext args, CommandSource source) throws CommandEx
 				throw new CommandException(args.getString(0) + " is not online.");
 			}
 		} else {
-			if (!(source instanceof VanillaPlayer)) {
+			if (!(source instanceof Player)) {
```
ZNickq, line 258: This should probably be VanillaPlayer, and you could change the message below to "you must be a vanilla player ..." :)
Other than that, looks fine!

windy1, line 258: No, this was a mistake on my part. It is right to be player. ;)

EDIT: To clarify, Player will never be a VanillaPlayer but rather player.getEntity().getController() will be.

ZNickq, line 258: Hmm? how? O: Isn't player the SpoutAPI player?
LE: oh, i failed -.- 
Hmm, get on irc, lawl

windy1, line 258: Think of player as an added level of abstraction basically. `SpoutPlayer implements Player` -> `Player.getEntity()` (Player HAS-AN Entity) -> `Entity.getController()` (Entity HAS-A Controller) (Would be a PlayerController in this case)

An entity can have a PlayerController as their controller (ie. VanillaPlayer) but Player is not actually a controller itself.

PR #130, opened 2012-04-15
Warning fix

TomyLobo:
some other minor changes included as well

### DIFF
--- a/src/main/java/org/spout/vanilla/generator/normal/decorator/MushroomDecorator.java
+++ b/src/main/java/org/spout/vanilla/generator/normal/decorator/MushroomDecorator.java
@@ -79,10 +79,12 @@ public class MushroomDecorator implements BiomeDecorator {
          */
     }
 
+    @SuppressWarnings("unused")
     private void generateMushroom(Chunk c, Random ra, int cx, int cy, int cz) {
         c.setBlockMaterial(cx, cy, cz, getRandomMushroomType(ra, false), (short) 0, true, c.getWorld());
     }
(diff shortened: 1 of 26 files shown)


ZNickq:
Squash the commits! Other than that, looks fine!

src/main/java/org/spout/vanilla/generator/normal/decorator/MushroomDecorator.java
```diff
@@ -79,11 +79,13 @@ public void populate(Chunk chunk, Random random) {
...
 
-    private void generateMushroom(Chunk c, Random ra, int cx, int cy, int cz) {
+    @SuppressWarnings("unused")
+	private void generateMushroom(Chunk c, Random ra, int cx, int cy, int cz) {
         c.setBlockMaterial(cx, cy, cz, getRandomMushroomType(ra, false), (short) 0, true, c.getWorld());
     }
 
-    private void generateHugeMushroom(World world, BlockMaterial type, Random ra, int x, int y, int z) {
+    @SuppressWarnings("unused")
+	private void generateHugeMushroom(World world, BlockMaterial type, Random ra, int x, int y, int z) {
```
zml2008: This formatting doesn't look right. Either this line or the lines around it should be corrected.

TomyLobo:
the first 2 commits are separate self-contained changes. why should i squash them?

TomyLobo:
ok i fixed the space thing. added a commit to change the whole file to tabs too

PR #131, opened 2012-04-16
Various Plant Changes

aPunch:
I've made some more changes to plants.

Added
-LilyPad
-SugarCane

Renamed
-NetherWartBlock to NetherWart (since it is in the block package, "Block" is redundant
-LongGrass to DeadBush

Changed/Fixed
-Flower
-Sapling
-TallGrass

There is still much work to be done with growth updates and block updates. Some of the blocks are not dropping items onUpdate properly, as well. I still have lots of work to do with plants but we're making progress! :)

### DIFF
--- a/src/main/java/org/spout/vanilla/material/block/DeadBush.java
+++ b/src/main/java/org/spout/vanilla/material/block/DeadBush.java
@@ -25,17 +25,21 @@
  */
 package org.spout.vanilla.material.block;
 
+import org.spout.api.Source;
+import org.spout.api.geo.World;
 import org.spout.api.material.Material;
+import org.spout.api.material.block.BlockFace;
 
 import org.spout.vanilla.material.Plant;
+import org.spout.vanilla.material.VanillaMaterials;
 import org.spout.vanilla.material.attachable.GroundAttachable;
 
-public class LongGrass extends GroundAttachable implements Plant {
-	public LongGrass(String name, int id) {
+public class DeadBush extends GroundAttachable implements Plant {
+	public DeadBush(String name, int id) {
 		super(name, id);
 	}
 
-	public LongGrass(String name, int id, int data, Material parent) {
+	public DeadBush(String name, int id, int data, Material parent) {
 		super(name, id, data, parent);
 	}
 
@@ -52,5 +56,14 @@ public class LongGrass extends GroundAttachable implements Plant {
 	@Override
 	public boolean hasGrowthStages() {
 		return false;
+	}
+
+	@Override
+	public boolean canPlace(World world, int x, int y, int z, short data, BlockFace against, Source source) {
+		if (super.canPlace(world, x, y, z, data, against, source)) {
+			return world.getBlock(x, y - 1, z).move(against.getOpposite()).getMaterial() == VanillaMaterials.SAND;
+		} else {
+			return false;
+		}
 	}
 }
--- a/src/main/java/org/spout/vanilla/material/block/LilyPad.java
+++ b/src/main/java/org/spout/vanilla/material/block/LilyPad.java
@@ -27,37 +27,18 @@ package org.spout.vanilla.material.block;
 
 import org.spout.api.Source;
 import org.spout.api.geo.World;
-import org.spout.api.geo.cuboid.Block;
 import org.spout.api.material.block.BlockFace;
-import org.spout.vanilla.material.Plant;
 import org.spout.vanilla.material.VanillaMaterials;
-import org.spout.vanilla.material.attachable.GroundAttachable;
 
-public class Flower extends GroundAttachable implements Plant {
-	public Flower(String name, int id) {
+public class LilyPad extends Solid {
+	public LilyPad(String name, int id) {
 		super(name, id);
-	}
-
-	@Override
-	public boolean hasGrowthStages() {
-		return false;
-	}
-
-	@Override
-	public int getNumGrowthStages() {
-		return 0;
-	}
-
-	@Override
-	public int getMinimumLightToGrow() {
-		return 8;
 	}
 
 	@Override
 	public boolean canPlace(World world, int x, int y, int z, short data, BlockFace against, Source source) {
 		if (super.canPlace(world, x, y, z, data, against, source)) {
-			Block block = world.getBlock(x, y, z).move(against.getOpposite());
-			return block.getMaterial() == VanillaMaterials.GRASS || block.getMaterial() == VanillaMaterials.DIRT;
+			return world.getBlock(x, y - 1, z).move(against.getOpposite()).getMaterial() == VanillaMaterials.WATER;
 		} else {
 			return false;
 		}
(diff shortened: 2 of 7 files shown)


src/main/java/org/spout/vanilla/material/block/DeadBush.java
```diff
@@ -53,4 +57,13 @@ public int getMinimumLightToGrow() {
 	public boolean hasGrowthStages() {
 		return false;
 	}
+
+	@Override
+	public boolean canPlace(World world, int x, int y, int z, short data, BlockFace against, Source source) {
+		if (super.canPlace(world, x, y, z, data, against, source)) {
+			return world.getBlock(x, y, z).move(against.getOpposite()).getMaterial() == VanillaMaterials.SAND;
```
ZNickq: This could allow someone to place a dead bush near sand, but over any other block...you probably need to check for y-1 to be sand :)

src/main/java/org/spout/vanilla/material/block/LilyPad.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class LilyPad extends Solid {
+	public LilyPad(String name, int id) {
+		super(name, id);
+	}
+
+	@Override
+	public boolean canPlace(World world, int x, int y, int z, short data, BlockFace against, Source source) {
+		if (super.canPlace(world, x, y, z, data, against, source)) {
+			return world.getBlock(x, y, z).move(against.getOpposite()).getMaterial() == VanillaMaterials.WATER;
```
ZNickq: Same...

src/main/java/org/spout/vanilla/material/block/DeadBush.java
aPunch: I don't really know what you mean and am unable to reproduce this.

ZNickq: ok, i do not have a computer to test on, but what happens if you try to place a dead bush while looking at a sand block, from north? It checks the sand block itself, but doesn't check what will be under the bush once it is placed! I will do a proper test tomorrow, since i will be finally home :P

aPunch: Assuming you mean from north as in the NORTH BlockFace, everything works as intended. The dead bush is able to be placed on sand and nothing else. Maybe I'm doing something wrong...

MWisBest: I think he's trying to say, what if you click a wall (vertical) of sand, and not a floor (horizontal) of sand. It appears your code is only checking to see if the block they click is sand, not the block where the dead bush ends up on (1 block underneath it, hence "y - 1").

aPunch: Ah, I see what you mean now. Fixing...

aPunch:
Fixed the canPlace issues. I've tested it somewhat thoroughly and the issue you brought up seems to be fixed. It would be nice if you could confirm it. :)

src/main/java/org/spout/vanilla/material/block/SugarCane.java
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		if (!validBases.contains(world.getBlockMaterial(x, y - 1, z))) {
+			Point point = new Point(world, x, y, z);
+			world.setBlockMaterial(x, y, z, VanillaMaterials.AIR, (short) 0, true, world);
+			world.createAndSpawnEntity(point, new Item(new ItemStack(VanillaMaterials.SUGAR_CANE, amount), point.normalize()));
+		}
+	}
+
+	public SugarCane initialize() {
```
Zidane: I think we already solved this issue in VanillaMaterials...can someone come along and confirm this for me.

@Olloth

aPunch: You solved a similar issue with drops by only setting drops after all materials have been initialized. Valid bases need to be set after all materials are initialized, am I correct?

src/main/java/org/spout/vanilla/material/block/SugarCane.java
```diff
@@ -0,0 +1,90 @@
...
+		} else {
+			return false;
+		}
+	}
+
+	@Override
+	public void onUpdate(World world, int x, int y, int z) {
+		int amount = 0;
+		int off = 1;
+		while (world.getBlockMaterial(x, y + off, z).equals(VanillaMaterials.CACTUS)) {
```
Zidane: I am confused to the cactus here...

src/main/java/org/spout/vanilla/material/block/SugarCane.java
Zidane: I think you may be correct on this but so long as SugarCane is initialized after the bases it should be fine (and thus this method is un-needed).

Order is important in VanillaMaterials.

aPunch: Removed the init method.

src/main/java/org/spout/vanilla/material/block/SugarCane.java
aPunch: Fixed.

src/main/java/org/spout/vanilla/material/block/TallGrass.java
```diff
@@ -25,7 +25,13 @@
...
 package org.spout.vanilla.material.block;
 
-public class TallGrass extends LongGrass {
+import org.spout.api.Source;
+import org.spout.api.geo.World;
+import org.spout.api.geo.cuboid.Block;
+import org.spout.api.material.block.BlockFace;
+import org.spout.vanilla.material.VanillaMaterials;
+
+public class TallGrass extends DeadBush {
```
Zidane, line 34: Last thing @aPunch and I swear I will shutup...

TallGrass is a child of DeadBush? That doesn't sound right...

TallGrass should be a child of Grass or you could make the hierarchy....

Fauna
- TallGrass
- DeadBush
- etc.

Zidane:
Also I want to mention @aPunch That @Afforess will be working on the growth system of Vanilla.

You two could perhaps collaborate.

src/main/java/org/spout/vanilla/material/block/TallGrass.java
aPunch, line 34: DeadBush was a child of TallGrass before I started on this. It was originally named LongGrass. You'll have to talk with whoever originally implemented TallGrass and LongGrass to understand the reasoning behind it.

windy1:
Growth might not be too difficult with our current set-up. Couldn't we just have plants have a BlockController, or is there something I'm missing?

Afforess:
Growth is a bit more complicated than that. I'll write up an outline of what growth and decay needs to do, and my idea of how to achieve it. In brief, growth needs to occur even in unloaded chunks - not just when players are nearby.

windy1:
Ah right hadn't thought about that, I guess I need to play more Minecraft :D

PR #134, opened 2012-04-21
Small patch for what appears to be simple mistakes in VanillaPlayer

VictorD:


ZNickq:
Thanks :)

PR #135, opened 2012-04-22
Pond generation code

DDoS:
Does't include populating code, I'll work on it after my exams.
The noise generation might need a bit of tweaking.

This time I've got the tabs and warnings right :D.


DDoS:
Since isOpaque() have been added to the SpoutAPI, I'll have to modify some stuff. Also, I have some optimization to do.

PR #136, opened 2012-04-22
Pond generation code

DDoS:
I haven't done the populating code because of my approaching finals, I'll get to it afterwards. I've tested the generation code with CraftBukkit and Spout, works nicely, but might need some very minor tweaking of the noise values.


ZNickq:
Well, looks fine :)

PR #142, opened 2012-04-27
Patch to let players drop items

VictorD:
This lets players drop items by pressing Q or throwing them outside their inventory.

Block collision is kind of temporary (and only in Y-axis) since I don't know how to use the real stuff and gravity is just -0.04f applied each tick, since GravityAction is meant for mobs(?) :)

### DIFF
--- a/src/main/java/org/spout/vanilla/controller/living/player/VanillaPlayer.java
+++ b/src/main/java/org/spout/vanilla/controller/living/player/VanillaPlayer.java
@@ -44,6 +44,7 @@ import org.spout.api.player.Player;
 import org.spout.vanilla.configuration.VanillaConfiguration;
 import org.spout.vanilla.controller.VanillaControllerTypes;
 import org.spout.vanilla.controller.living.Human;
+import org.spout.vanilla.controller.object.moving.Item;
 import org.spout.vanilla.controller.source.HealthChangeReason;
 import org.spout.vanilla.protocol.msg.DestroyEntityMessage;
 import org.spout.vanilla.protocol.msg.PingMessage;
@@ -218,6 +219,9 @@ public class VanillaPlayer extends Human implements PlayerController {
 		}
 
 		inventory.setCurrentSlot(0);
+		if (getActiveInventory() == null)
+			setActiveInventory(inventory);
+
 		return inventory;
 	}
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/spout/vanilla/protocol/handler/DiggingMessageHandler.java
```diff
@@ -97,6 +99,10 @@ public void handleServer(Session session, Player player, DiggingMessage message)
 		} else if (message.getState() == DiggingMessage.STATE_DONE_DIGGING) {
 			//TODO: Timing checks!
 			blockBroken = true;
+		} else if (message.getState() == DiggingMessage.STATE_DROP_ITEM && isVanilla) {
+			VanillaPlayer vplayer = (VanillaPlayer) player.getEntity().getController();
+			if (vplayer != null)
```
Top-Cat, line 104: OTB add the brackets

Few other places too

src/main/java/org/spout/vanilla/controller/living/player/VanillaPlayer.java
```diff
@@ -492,4 +496,83 @@ public ItemStack getItemOnCursor() {
...
+
+		ItemStack dropStack = stack.clone();
+		dropStack.setAmount(amount);
+		Item newItem = new Item(dropStack, velocity);
+		Vector3 position = e.getPosition().add(itemPositionOffset);
+		e.getWorld().createAndSpawnEntity(new Point(position, getParent().getWorld()), newItem);
+		return newItem;
+	}
+
+	private static ItemStack subtractFromStack(ItemStack in, int count) {
```
ZNickq, line 567: This shouldn't be in VanillaPlayer,more like SpoutAPI...

src/main/java/org/spout/vanilla/controller/living/player/VanillaPlayer.java
```diff
@@ -218,6 +219,9 @@ public PlayerInventory createInventory(int size) {
 		}
 
 		inventory.setCurrentSlot(0);
+		if (getActiveInventory() == null)
```
ZNickq, line 222: Pretty sure createInventory sets the inventory immediately afterwards...was there an issue with this?

VictorD, line 222: Ah I see. Without that line I couldn't press Q after joining and grabbing an item but had to open the inventory, lift it up and put it down again so I figured it wasn't set

src/main/java/org/spout/vanilla/protocol/handler/DiggingMessageHandler.java
VictorD, line 104: OTB? I assume it means Omitting The Brackets or something , I've added them locally now

Top-Cat, line 104: No, one true brace is the formatting style

PR #143, opened 2012-04-29
Fix for entity head yaw http://issues.spout.org/browse/VANILLA-30

VictorD:
This patch modifies Living.java so that it updates the head yaw for living entities correctly


ZNickq:
Thanks!

PR #146, opened 2012-04-30
Minor inventory fixes

VictorD:
Added tmpvar to stack swapping and made stack splitting round uneven stacks correctly again


VictorD:
I see now that right click stack swapping is broken even with this patch, couldn't see that issue before since stack swapping was broken altogether oh well

windy1:
Thanks I've been working on inventory handling and still working out the blips. :)

PR #153, opened 2012-05-09
Implemented the smooth populator

DDoS:
It's gonna need more tweaking, but that's hard to do without proper biomes, and currently the world generated seems to be only one large biome.

It's a start at least.

### DIFF
--- a/src/main/java/org/spout/vanilla/world/generator/normal/biome/BeachBiome.java
+++ b/src/main/java/org/spout/vanilla/world/generator/normal/biome/BeachBiome.java
@@ -30,10 +30,11 @@ import org.spout.api.util.cuboid.CuboidShortBuffer;
 
 import org.spout.vanilla.material.VanillaMaterials;
 import org.spout.vanilla.world.generator.VanillaBiome;
+import org.spout.vanilla.world.populator.SmoothPopulator;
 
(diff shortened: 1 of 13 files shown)


zml2008:
imo the smoothing populator should be implemented as a normal populator not connected to a specific biome, run before the BiomeDecorators. This can be done within BiomeGenerator.

DDoS:
I was told to do it like this.

zml2008:
When we discussed this a few months ago we decided on the layout I described in my original comment, and I don't remember that decision changing. This is also where biome smoothing is placed in http://forums.spout.org/threads/a-look-at-vanilla-world-generation.1780/

Zidane:
I thought this was the best approach to it...didn't even think about doing it within the BiomeGenerator. @zml2008 is on to something here.

Zidane:
Yeah I admit this was my bad...I completely forgot about that. Thanks for letting us know @zml2008

DDoS:
Well, I guess I better close this.

Zidane:
@DDoSQc you can still adapt this easily to the BiomeGenerator :).

DDoS:
Yeah, once I'm done fighting my way through git, I'll have make a PR. I also added a addPopulator method to the BiomeGenerator.

PR #157, opened 2012-05-10
Changed the dungeon decorator to use the WorldGeneratorObject system

DDoS:
This is what I'm planning to do for most of the decorators, following a suggestion from @RoyAwesome
If I've got the team's approval, that's how I'll setup all the objects.


zml2008:
In my opinion settings should be controlled through setter methods in the object itself, instead of one big constructor with a ton of parameters. While they're a bit less compact, setter methods make the code a bunch more readable without referencing the original object.

DDoS:
Alright, noted.

PR #159, opened 2012-05-10
Fixes issue VANILLA-71

joshlit:
Fixed the issue with /give <itemnamehere> not assigning the item data to the given item.

Link to issue http://issues.spout.org/browse/VANILLA-71

Signed-off-by: Josh Latimer joshgesh@gmail.com


windy1:
This probably needs to be tested by a team member before the point can be awarded on spout.org, I would do it myself, but my rig's psu is failing again :/

PR #165, opened 2012-05-13
Safe spawn point implementation & WGO spires

DDoS:
Better than spawning in mid air.

Done converting the decorators, now I think I'll try to generate some basic nether terrain.


Zidane:
Want to improve the plains/etc biomes first? The normal world is a bit higher priority than the nether :P. Anyways...thanks as always :D.

PR #172, opened 2012-05-16
Unfinished implementation of savable extra item data, for maps and books!

ZNickq:
This is not finished, still need to actually add behaviour for the map, but i need to get some feedback on this, since i never used SpoutNBT (or NBT at all) before!

@Olloth @Raphfrk You guys know how to use it apparently :D


Salynne:
Vanilla does not and will not manually save or load anything from NBT like this. Totally wrong. ItemStacks need datatables and then they need to save themselves. (This goes right in the API)

The only reason NBT is even used in Vanilla, is because you need to send that over the protocol to vanilla clients.

PR #174, opened 2012-05-17
Improved the nether

DDoS:
256 blocks high and less patterned.

### DIFF
--- a/src/main/java/org/spout/vanilla/world/generator/nether/biome/NetherrackBiome.java
+++ b/src/main/java/org/spout/vanilla/world/generator/nether/biome/NetherrackBiome.java
@@ -27,6 +27,7 @@
 package org.spout.vanilla.world.generator.nether.biome;
 
 import net.royawesome.jlibnoise.NoiseQuality;
+import net.royawesome.jlibnoise.module.combiner.Displace;
 import net.royawesome.jlibnoise.module.modifier.Clamp;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/world/generator/nether/biome/NetherrackBiome.java
```diff
@@ -88,14 +110,12 @@ public NetherrackBiome(int id) {
...
+		mainDisplacer.setSeed(seed);
 		main.setSeed(seed);
 		bedrockBase.setSeed(seed);
-		if (chunkY > 7) {
-			return;
-		}
 		final int y = chunkY * 16;
 		for (int yy = y; yy < y + 16; yy++) {
-			if (yy == 0 || yy == 127) {
+			if (yy == 0 || yy == 255) {
```
windy1, line 118: use World.getMaxHeight() for y axis :)

Zidane:
Gonna pull this and then make the change to the nether myself (as it is something silly).

src/main/java/org/spout/vanilla/world/generator/nether/biome/NetherrackBiome.java
DDoS, line 118: If you're gonna do that, use world.getMaxHeight() - 1 and don't forget to change lines 130,131,143,144

PR #175, opened 2012-05-17
Initial Enchantment Table Implementation, part 2

aPunch:
This pull adds enchantment tables. I have also cleaned up some code all around, but mainly in the controller/controller block classes. I'd like to get this in before I begin work on enchantments.

### DIFF
--- a/src/main/java/org/spout/vanilla/inventory/EnchantmentTableInventory.java
+++ b/src/main/java/org/spout/vanilla/inventory/EnchantmentTableInventory.java
@@ -27,62 +27,79 @@
 package org.spout.vanilla.inventory;
 
 import org.spout.api.inventory.Inventory;
-
-import org.spout.vanilla.controller.block.ChestController;
+import org.spout.api.inventory.ItemStack;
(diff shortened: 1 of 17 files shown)


src/main/java/org/spout/vanilla/inventory/EnchantmentTableInventory.java
```diff
@@ -0,0 +1,104 @@
...
+		addViewer(player.getPlayer().getNetworkSynchronizer());
+		player.setActiveInventory(this);
+		player.openWindow(Window.ENCHANTMENT_TABLE, getSize());
+	}
+
+	@Override
+	public void onClosed(VanillaPlayer player) {
+		if (hasItem()) {
+			// Drop the item
+			player.getParent().getWorld().createAndSpawnEntity(player.getHeadPosition(), new Item(getItem(0), Vector3.FORWARD));
```
ZNickq: Also, remove the item from the table's inventory ;)

ZNickq:
Other than that, it seems right, also i have no idea how enchantment tables work ;)

PR #176, opened 2012-05-17
Material Cleanup and Small Refactor

aPunch:
-VanillaMaterials cleanup:
     -Remove unnecessary casts
     -Renamed loadProperties() to initialize()
     -Moved some calls to the initialize method

-Some small changes to RedstoneLamp, RedstoneRepeater, and other blocks that turn on/off

-Adds Pumpkin class

-Removed empty material constructors in favor of more standardized constructors


windy1:
This failed because of a casing issue with git, it doesn't seem to pick up casing changes in file names because when you renamed NetherRack to Netherrack it didn't pick up the file name change :/

aPunch:
Ah, that issue has bugged me before. Sorry that I didn't think of that before.

PR #180, opened 2012-05-17
Smooth populator fix

DDoS:
Make sure to stop checking for biome variations once it's been found to have some (so, stop using the last variation check, which might not represent the entire area).

Terrain generation looks a bit better now.
:D

Yeah, it's not a listed issue for the feature freeze, but it's an important fix.

Also, added a missing license header.

I've just completed the HugeTreeObject, should I make a PR, or wait for the end of the feature freeze?

### DIFF
--- a/src/main/java/org/spout/vanilla/material/block/solid/Pumpkin.java
+++ b/src/main/java/org/spout/vanilla/material/block/solid/Pumpkin.java
@@ -1,3 +1,29 @@
+/*
+ * This file is part of Vanilla.
+ *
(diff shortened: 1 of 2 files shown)


Zidane:
WGEN isn't scheduled for any milestone...you can continue to PR it.

Zidane:
The smooth populator would work so much better with some proper normal biomes @DDoSQc ;D

DDoS:
@NinjaZidane Yes, I know, do the biomes, you don't have to be subtle about it. They're not easy, generation is a combination of 3D noise using density and 2D using height, which I am unsure how to do. I guess I'll have to take a closer look at the mc source code.

PR #183, opened 2012-05-18
Implemented Enchantments

aPunch:
This is Part 2 of the enchantment system implementation. I've implemented all vanilla enchantments in a similar way to how VanillaMaterials are handled. The next step is attaching enchantments to ItemStacks and making the enchantment table work!

@W1ndwaker Are the three bars in the enchantment table considered slots? How are they handled?


windy1:
Thank you, and no, the bars are not counted as slots. Instead another packet is sent to the server when a player enchants an item. http://wiki.vg/Protocol#Enchant_Item_.280x6C.29 ;)

PR #185, opened 2012-05-19
Init fixes

Owexz:
(Notably Opacity) Signed off by: Owexz Owexz@Owexz.net
Signed off by: Owexz Owexz@Owexz.net

Here you go Zidane


stephenwilliams:
Had to be reverted: your Signed-off-by was invalid http://forums.spout.org/wiki/pull-request-guidelines/

windy1:
Your sign off was incorrect, so instead of fixing the compile mistakes I had to remove the commit. I'm sorry for the trouble, and I'm sorry that I did not notice but for your sign off to be valid it needs to be in the form of.
`Signed-off-by: Owexz <example@spout.org>`

PR #187, opened 2012-05-20
VanillaItemStack and Silk Touch Functionality

aPunch:
This pull adds VanillaItemStack to use for enchantment-related methods (and other Vanilla-specific methods as the need arises). I have also implemented drop rules for blocks that are affected by the Silk Touch enchantment. Some may be incorrect, but I tried my best to get them all correct the first time.


Zidane:
Manually merged in by me

PR #190, opened 2012-05-20
Update ItemStack References to VanillaItemStack

aPunch:
This should fix and prevent any ClassCastExceptions in the future. Now all items in Vanilla will be VanillaItemStacks.


aPunch:
This breaks more than it fixes. Closing.

PR #191, opened 2012-05-20
Small fixes for the smooth populator and trees

DDoS:
Nothing big...


windy1:
Actually, I'm kind of curious as to why you made TreeObject protected

DDoS:
The class is abstract, you can't initialize it, the constructor is only used by the subclass.

windy1:
Ah, I hadn't even noticed :p

PR #193, opened 2012-05-22
Improved static initialisation order test to log more detailed errors in some cases

LunNova:
In cases where the issue is caused by failed static initialisation order inside a material class - instead of in VanillaMaterials - more detail is now given.


windy1:
:'(

PR #196, opened 2012-05-23
Fixed formatting. Added Xlint compiler arguments and show warnings/depreciation. Fixed issues found with lint warnings.

LunNova:
Fixed bad mob spawner logic - spawns creepers if not in range for silverfish?!

Fixed uncheckd conversions. Removed useless casts.

Undid some of windwaker's formatting - using latest Spout.xml here so @w1ndwaker's must be the incorrect formatting, unless I messed something up.

I may have too many things in this PR. Also, need to know if the formatting here is actually correct, and if not, how do I fix IDEA not to do this?

### DIFF
--- a/Spout.xml
+++ b/Spout.xml
@@ -64,5 +64,10 @@
       <option name="USE_TAB_CHARACTER" value="true" />
     </indentOptions>
   </codeStyleSettings>
+  <codeStyleSettings language="XML">
+    <indentOptions>
+      <option name="USE_TAB_CHARACTER" value="true" />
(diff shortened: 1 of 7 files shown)


Zidane:
As much as I appreciate PRs, I will have to decline this as I never allow formatting ones. We get around to fixing the formatting in Vanilla based on what we use. If Windwaker's is wrong I will take another look at it to see what is up.

If you wanna strip the formatting from it and make a PR containing your other fixes, I will gladly pull.

PR #197, opened 2012-05-24
Silk Touch Enchantments

aPunch:
This re-adds all of the silk touch effects that were reverted before.


src/main/java/org/spout/vanilla/material/block/controlled/SilverfishStone.java
```diff
@@ -24,19 +24,19 @@
...
 	public SilverfishStone(String name, int id) {
 		super(name, id);
-		this.setController(VanillaControllerTypes.SILVERFISH_STONE);
 	}
 
 	@Override
 	public void initialize() {
 		super.initialize();
-		this.setHardness(0.8F).setResistance(1.2F); //TODO Confirm that this is the correct value
+		this.setHardness(0.8F).setResistance(1.2F).setController(VanillaControllerTypes.SILVERFISH_STONE); // TODO Confirm that this is the correct value
```
Zidane: Should be in the constructor I believe (setController).

aPunch: Well, there were some setController's in the constructor and some in the init method, so I moved them all to the init for consistency. I'll move them to the constructor, now.

EDIT - Done.

windy1:
The getDrops method has a lot of duplicate code, we can probably abstract at least some of it, no?

aPunch:
@W1ndwaker Not anymore than what was in the getDrops before. There are so many use-cases that creating more helper methods in EnchantmentUtil would seem like a waste unless everything was _really_ generalized. And even then I would think there would be duplicated code. I'll see what I can do to cut down on it, though.

windy1:
@aPunch Yeah I know it wasn't you just stating it, if you could do something about it that would've awesome :)

LunNova:
@W1ndwaker 

I'm working on that.

PR #206, opened 2012-05-26
Added sensible default getDrops implementation to reduce code duplication.

LunNova:
Note: code duplication not actually removed yet, this just makes it possible.

### DIFF
--- a/src/main/java/org/spout/vanilla/material/VanillaBlockMaterial.java
+++ b/src/main/java/org/spout/vanilla/material/VanillaBlockMaterial.java
@@ -26,6 +26,7 @@
  */
 package org.spout.vanilla.material;
 
+import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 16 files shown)


Zidane:
rebase for me @nallar

LunNova:
I thought we decided in #Vanilla that this way didn't allow enough and would still require overriding getDrops in too many cases, and I still haven't finished the other way.

Zidane:
Erm that is correct @nallar ...want to close this or still use it?

windy1:
I don't really see any other way, if you need to add on to this you could just override it and make sure you call super#getDrops

LunNova:
@W1ndwaker 

This doesn't allow for multiple drops or random numbers of drops, and that would be possible.

windy1:
Alright, I'm going to go ahead and close it then.

PR #207, opened 2012-05-26
Change to getDrops in VanillaActionController

aPunch:

This pull changes getDrops to allow for source-based drops, such as cooked meat if a chicken was last damaged by fire. It also adds a DamageCause source for possible damage causes. Some HealthChangeReason's were moved to DamageCause. A new "DAMAGE" HealthChangeReason has been added that points to DamageCause.

In the second commit, I made changes to enchantability and added SMITE, BANE_OF_ARTHROPODS, AND SHARPNESS. Zidane wanted separate commits for different functions.

### DIFF
--- a/src/main/java/org/spout/vanilla/controller/living/creature/hostile/Creeper.java
+++ b/src/main/java/org/spout/vanilla/controller/living/creature/hostile/Creeper.java
@@ -29,13 +29,16 @@ package org.spout.vanilla.controller.living.creature.hostile;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.spout.api.Source;
 import org.spout.api.entity.type.ControllerType;
 import org.spout.api.entity.type.EmptyConstructorControllerType;
 import org.spout.api.inventory.ItemStack;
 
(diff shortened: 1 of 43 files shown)


src/main/java/org/spout/vanilla/controller/living/creature/hostile/Creeper.java
```diff
@@ -54,14 +57,16 @@ public void onAttached() {
 	}
 
 	@Override
-	public Set<ItemStack> getDrops() {
+	public Set<ItemStack> getDrops(Source source, VanillaActionController lastDamager) {
```
ZNickq, line 60: I don't like using VanillaActionController here, why not give it the EntityHealthChangeEvent?

aPunch, line 60: Why would I pass the event to getDrops when all I need to know is when the entity dies?

Zidane, line 60: What is the difference between source and lastDamager here @aPunch 

EDIT: I see now.

PR #209, opened 2012-05-26
Hackish pom.xml fix - previous syntax is invalid XML.

LunNova:
However Maven handles it fine. Unfortunately, IDEA is more picky. This fix works as the "-O" option is a legacy option which javac ignores, instead of throwing an error, and then parses the next part of that argument. May break with future compilers.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,7 +174,7 @@
 					<target>1.6</target>
 					<compilerArguments>
 						<Xlint:all/>
-						<Xlint:-path/>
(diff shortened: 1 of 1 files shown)


LunNova:
And the reason to do this is to solve the "OMG I CAN'T IMPORT VANILLA TO IDEA?!" issue.

lukespragg:
I'm not certain that this is the best solution for this. I'll pull it for now, but will likely find a cleaner way in my next pom update.

PR #214, opened 2012-05-28
Implemented /Help Command

fallensk8ur:
This lists all Vanilla commands by using command /Help

Signed-off-by: Fallensk8ur fallensk8ur1234@aol.com


zml2008:
If you're going to do a help command, make it construct the help listing from all registered commands and put it in Spout.

fallensk8ur:
So, you want me to add these Vanilla commands in the help menu on Spout as well?

PR #215, opened 2012-05-28
Respiration Enchantment and First Signs of Survival

aPunch:

To add the respiration enchantment, I needed drowning and suffocation to be implemented. Of course, the damage system will need to be expanded to account for things like armor, potion effects, etc., but for now this pull introduces some survival-like behavior for people to try out.

### DIFF
--- a/src/main/java/org/spout/vanilla/controller/VanillaActionController.java
+++ b/src/main/java/org/spout/vanilla/controller/VanillaActionController.java
@@ -68,6 +68,7 @@ public abstract class VanillaActionController extends ActionController implement
 	private int fireTicks = 0;
 	private int positionTicks = 0;
 	private int velocityTicks = 0;
+	protected int airTicks = 0;
 	// Velocity-related
 	private Vector3 velocity = Vector3.ZERO;
 	private Vector3 movementSpeed = Vector3.ZERO;
@@ -100,6 +101,7 @@ public abstract class VanillaActionController extends ActionController implement
 		maxSpeed = (Vector3) data().get("max_speed", maxSpeed);
 		health = (Integer) data().get("health", maxHealth);
 		maxHealth = (Integer) data().get("max_health", maxHealth);
+		airTicks = (Integer) data().get("air_ticks", airTicks);
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


src/main/java/org/spout/vanilla/controller/VanillaActionController.java
```diff
@@ -298,26 +301,15 @@ public void damage(int amount, DamageCause cause) {
...
 	 * @param cause cause of this controller being damaged
 	 * @param damager controller that damaged this controller
-	 * @param sendHurtMessage whether or not to send a hurt message
 	 */
-	public void damage(int amount, DamageCause cause, VanillaActionController damager, boolean sendHurtMessage) {
+	public void damage(int amount, DamageCause cause, VanillaActionController damager) {
 		setHealth(getHealth() - amount, new HealthChangeReason(HealthChangeReason.Type.DAMAGE));
 		lastDamager = damager;
-		if (sendHurtMessage) {
+		if (amount > 0) {
```
windy1: I believe this message still gets sent even if the player is not damaged. Take snowballs for example.

aPunch: I've always thought that playing the hurt animation for zero damage was a bug or an oversight. :P

windy1: How would the opponent know if the snowball hit or not? :p

src/main/java/org/spout/vanilla/controller/living/player/VanillaPlayer.java
```diff
@@ -81,7 +84,8 @@
 	protected Window activeWindow = new DefaultWindow(this);
 	protected String tabListName;
 	protected GameMode gameMode;
-	protected int distanceMoved, miningDamagePeriod = VanillaConfiguration.PLAYER_SPEEDMINING_PREVENTION_PERIOD.getInt(), miningDamageAllowance = VanillaConfiguration.PLAYER_SPEEDMINING_PREVENTION_ALLOWANCE.getInt();
+	protected int distanceMoved, miningDamagePeriod = VanillaConfiguration.PLAYER_SPEEDMINING_PREVENTION_PERIOD.getInt(),
```
windy1, line 87: Line break :/

PR #218, opened 2012-05-29
Crafting inventories/windows

kitskub:
This sets up Crafting Tables for crafting. All it needs now is work on the recipe side and searching through recipes and updating accordingly. Feel more than free to make comments/scream at how I messed up on something and the like.

Related: https://github.com/SpoutDev/SpoutAPI/pull/233'
This also partially fixes: http://issues.spout.org/browse/VANILLA-72

### DIFF
--- a/src/main/java/org/spout/vanilla/inventory/CraftingTableInventory.java
+++ b/src/main/java/org/spout/vanilla/inventory/CraftingTableInventory.java
@@ -24,51 +24,22 @@
  * License and see <http://www.spout.org/SpoutDevLicenseV1.txt> for the full license,
  * including the MIT license.
  */
-package org.spout.vanilla.material.block.solid;
-
-import java.util.ArrayList;
-
-import org.spout.api.entity.Controller;
-import org.spout.api.entity.Entity;
-import org.spout.api.event.player.PlayerInteractEvent.Action;
-import org.spout.api.geo.cuboid.Block;
-import org.spout.api.inventory.ItemStack;
-import org.spout.api.material.block.BlockFace;
+package org.spout.vanilla.inventory;
 
 import org.spout.vanilla.controller.living.player.VanillaPlayer;
-import org.spout.vanilla.inventory.CraftingTableInventory;
-import org.spout.vanilla.material.block.Solid;
-import org.spout.vanilla.util.Instrument;
+import org.spout.vanilla.window.CraftingWindow;
+import org.spout.vanilla.window.Window;
 
-public class CraftingTable extends Solid {
+public class CraftingTableInventory extends WindowInventory{
+	private static final long serialVersionUID = 1L;
 	
-	public CraftingTable(String name, int id) {
-		super(name, id);
-		this.setHardness(4.2F);
-	}
-		
-	@Override
-	public Instrument getInstrument() {
-		return Instrument.BASSGUITAR;
-	}
-
-	@Override
-	public ArrayList<ItemStack> getDrops(Block block) {
-		ArrayList<ItemStack> drops = new ArrayList<ItemStack>();
-		drops.add(new ItemStack(this, 1));
-		return drops;
+	public CraftingTableInventory() {
+		super(9);
 	}
 	
 	@Override
-	public void onInteractBy(Entity entity, Block block, Action action, BlockFace face) {
-		if (action == Action.RIGHT_CLICK) {
-			Controller controller = entity.getController();
-			if (!(controller instanceof VanillaPlayer)) {
-				return;
-			}
+	public Window createWindow(VanillaPlayer player) {
+		return new CraftingWindow(player, this);
 
-			// Open the crafting table
-			(new CraftingTableInventory()).open((VanillaPlayer) controller);
-		}
 	}
 }
--- a/src/main/java/org/spout/vanilla/window/CraftingWindow.java
+++ b/src/main/java/org/spout/vanilla/window/CraftingWindow.java
@@ -26,10 +26,65 @@
  */
 package org.spout.vanilla.window;
 
+import org.spout.api.inventory.ItemStack;
 import org.spout.vanilla.controller.living.player.VanillaPlayer;
+import org.spout.vanilla.controller.object.moving.Item;
+import org.spout.vanilla.inventory.CraftingTableInventory;
 
 public class CraftingWindow extends Window {
-	public CraftingWindow(VanillaPlayer owner) {
+	private static final int SLOTS[] = {37, 38, 39, 40, 41, 42, 43, 44, 45, 28, 29, 30, 31, 32, 33, 34, 35, 36, 19, 20, 21, 22, 23, 24, 25, 26, 27, 10, 11, 12, 13, 14, 15, 16, 17, 18, 7, 8, 9, 4, 5, 6, 1, 2, 3, 0};
+	
+	public CraftingWindow(VanillaPlayer owner, CraftingTableInventory craftingInventory) {
 		super(1, "Crafting", owner);
+		this.setInventory(owner.getInventory().getItems(), craftingInventory);
+		this.setSlotConversionArray(SLOTS);
 	}
+
+	@Override
+	public void onClosed() {
+		for (int slot = 0; slot < this.getInventory().getSize() - 9; slot++) {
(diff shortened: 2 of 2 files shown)


src/main/java/org/spout/vanilla/material/block/controlled/CraftingTableBlock.java
```diff
@@ -0,0 +1,81 @@
...
+import org.spout.api.inventory.ItemStack;
+import org.spout.api.material.block.BlockFace;
+
+import org.spout.vanilla.controller.VanillaControllerTypes;
+import org.spout.vanilla.controller.block.CraftingTable;
+import org.spout.vanilla.controller.living.player.VanillaPlayer;
+import org.spout.vanilla.inventory.CraftingTableInventory;
+import org.spout.vanilla.util.Instrument;
+
+public class CraftingTableBlock extends ControlledMaterial {
```
windy1: Should just be "CraftingTable" the "Block" is redundant

aPunch:
Why does CraftingTable need to be controlled? Can't recipes be checked for in CraftingTableInventory's onClick methods?

windy1:
It doesn't and I said the same thing in IRC @Kitskub

src/main/java/org/spout/vanilla/material/block/controlled/CraftingTableBlock.java
kitskub: If it's "CraftingTable" then there are a ton of errors because its the same name as the controller, so this is what zidane told me to do.

kitskub:
Unfortunately, it does for the same reason that Enchantment table needs a controller: it holds the instance of an inventory. However, if there's another way to do this, please point it out.

aPunch:
CraftingTable shouldn't need to hold the instance of any inventory. All it needs is an onInteractBy method that points to a new CraftingTableInventory, which will handle all crafting.

src/main/java/org/spout/vanilla/material/block/controlled/CraftingTableBlock.java
LunNova: Use the full name when referencing org.spout.vanilla.controller.block.CraftingTable.

windy1:
@aPunch is right @kitskub the enchantment table will be used for other data besides the inventory.

kitskub:
Ok, but it actually is going to need to be a map of player to inventory, because as it is, all players see the same one.

windy1:
No, you would create a new instance of the inventory every CraftingTable#onInteract

kitskub:
Yeah, just looked through the code, and that'll work.

kitskub:
Fixed

src/main/java/org/spout/vanilla/material/block/solid/CraftingTable.java
```diff
@@ -1,54 +1,75 @@
...
+import org.spout.api.inventory.ItemStack;
+import org.spout.api.material.block.BlockFace;
+
+import org.spout.vanilla.controller.living.player.VanillaPlayer;
+import org.spout.vanilla.inventory.CraftingTableInventory;
+import org.spout.vanilla.material.block.Solid;
+import org.spout.vanilla.util.Instrument;
+
+public class CraftingTable extends Solid {
+	CraftingTableInventory inventory;
```
windy1: No, we don't want a global field, we want to instantiate an entirely new CraftingTableInventory every CraftingTable#onInteract

(edit) s/global field/instance variable
(edit) oh.. wait... why is this here?

src/main/java/org/spout/vanilla/window/CraftingWindow.java
```diff
@@ -26,10 +26,67 @@
...
+	
+	public CraftingWindow(VanillaPlayer owner, CraftingTableInventory craftingInventory) {
 		super(1, "Crafting", owner);
+		this.setInventory(owner.getInventory().getItems(), craftingInventory);
+		this.setSlotConversionArray(SLOTS);
 	}
+
+	@Override
+	public void onClosed() {
+		for (int slot = 0; slot < this.getInventory().getSize() - 9; slot++) {
```
windy1, line 45: What is the point of this loop?

src/main/java/org/spout/vanilla/window/CraftingWindow.java
```diff
@@ -26,10 +26,67 @@
...
+			ItemStack current = this.getInventory().getItem(getSpoutSlotIndex(slot));
+			if (current == null) {
+				continue;
+			}
+			Item control = new Item(current, getOwner().getHeadPosition().add(getOwner().getLookingAt()));
+			getOwner().getParent().getWorld().createAndSpawnEntity(getOwner().getHeadPosition().add(0.0, -0.4, 0.0), control);
+			this.getInventory().setItem(getSpoutSlotIndex(slot), null);
+		}
+        }
```
windy1: Formatting

windy1:
Also, slot conversion should be bottom to top and left to right.

src/main/java/org/spout/vanilla/material/block/solid/CraftingTable.java
kitskub: Totally missed this when I got rid of it. Oops.

stephenwilliams: Its not global, it just needs to be change to private, right now its access level is default

src/main/java/org/spout/vanilla/material/block/solid/CraftingTable.java
```diff
@@ -1,54 +1,75 @@
...
+	@Override
+	public void onInteractBy(Entity entity, Block block, Action action, BlockFace face) {
+		if (action == Action.RIGHT_CLICK) {
+			Controller controller = entity.getController();
+			if (!(controller instanceof VanillaPlayer)) {
+				return;
+			}
+
+			// Open the crafting table
+			(new CraftingTableInventory()).open((VanillaPlayer) controller);
```
windy1, line 71: Even though this is correct I would prefer

``` java
CraftingTableInventory inventory = new CraftingTableInventory();
inventory.open((VanillaPlayer) controller);
```

src/main/java/org/spout/vanilla/window/CraftingWindow.java
kitskub, line 45: To drop all the items when the craftingwindow is closed

src/main/java/org/spout/vanilla/window/CraftingWindow.java
```diff
@@ -26,10 +26,67 @@
...
+	
+	@Override
+	public boolean onLeftClick(int clickedSlot) {
+		if (clickedSlot == 0) {
+			if (this.hasItemOnCursor()) {
+				return false;
+			}
+			else {
```
windy1: Your else statement is not formatted correctly; needs to be 

``` java
if (true) {
    // something
} else {
    // something else
}
```

not

``` java
if (true) {
    // something
}
else {
    // something else
}
```

src/main/java/org/spout/vanilla/window/CraftingWindow.java
```diff
@@ -26,10 +26,67 @@
...
+		return true;
+	}
+	
+	@Override
+	public boolean onRightClick(int clickedSlot) {
+		if (clickedSlot == 0) {
+			if (this.hasItemOnCursor()) {
+				return false;
+			}
+			else {
```
windy1: same here

src/main/java/org/spout/vanilla/inventory/CraftingTableInventory.java
```diff
@@ -0,0 +1,47 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License,
+ * the MIT license and the SpoutDev License Version 1 along with this program.
+ * If not, see <http://www.gnu.org/licenses/> for the GNU Lesser General Public
+ * License and see <http://www.spout.org/SpoutDevLicenseV1.txt> for the full license,
+ * including the MIT license.
+ */
+package org.spout.vanilla.inventory;
+
```
windy1, line 28: Two lines of blank code MAX

kitskub:
Should ALL be fixed
Except the slot stuff should be double checked as I honestly just am guessing at that.

PR #219, opened 2012-05-29
Damage System

aPunch:

This adds a better damage system. It takes into account the player's armor and whether the armor is enchanted. In the future it can be expanded upon to include potion effects.

I agree that this needs to be changed/fleshed out more, so provide any feedback that you have.


src/main/java/org/spout/vanilla/controller/living/player/VanillaPlayer.java
```diff
@@ -246,7 +246,7 @@ private void updateHealthAndHunger() {
 			System.out.println("Hunger: " + hunger);
 			System.out.println("Health: " + health);
 			System.out.println("Exhaustion: " + exhaustion);
-			sendPacket(owner, new UpdateHealthMessage(health, hunger, foodSaturation));
+			// sendPacket(owner, new UpdateHealthMessage(health, hunger, foodSaturation));
```
windy1, line 249: Any reason you commented this out?

LunNova, line 249: I think there was an issue with it resulting in the hurt animation.

aPunch, line 249: I moved the UpdateHealthMessage to setHealth in VanillaPlayer. Also, there is an issue with regeneration where the hurt animation is played, but apparently that cannot be fixed in an easy manner until SpoutClient is ready. Am I correct on this, @NinjaZidane?

windy1, line 249: That can't be, it was working fine before. The rest of this PR looks fine so I'll go ahead and pull it and look into this later today. :)

PR #222, opened 2012-05-31
Rewrite of recipes

kitskub:
This is a rewrite of recipes. This uses a builder pattern, and makes recipes immutable. Example implementations are in VanillaRecipes, but because @Olloth wants to make them parsed from a file, I didn't put all of them. I welcome all commentary.

Corresponding SpoutAPI Pull: https://github.com/SpoutDev/SpoutAPI/pull/235
One issue: http://issues.spout.org/browse/SPOUTAPI-15


kitskub:
Updated to include a RecipeBuilderTest. I can add more tests if its wanted.

PR #224, opened 2012-06-01
VANILLA-127

aPunch:

This fixes VANILLA-127 where breaking cake (and many other blocks) would kick the player because it did not implement Mineable.


Darsstar:
The cake is a LIE!

src/main/java/org/spout/vanilla/material/block/plant/Cactus.java
```diff
@@ -117,4 +120,24 @@ public float getCraftTime() {
...
+	}
+
+	@Override
+	public boolean hasGrowthStages() {
+		return true;
+	}
+
+	@Override
+	public int getNumGrowthStages() {
+		return 3;
```
stephenwilliams, line 136: Why not make stuff like this modifiable?

aPunch, line 136: As of right now, the growth system is not even implemented yet. I think @Afforess is working on that, and he can decide how modifiable he wants it to be.

PR #232, opened 2012-06-04
Change to CakeBlock

aPunch:

Having sub materials for cake blocks seems unnecessary, so I added a CakeSize enum to CakeBlock and set its data instead.


ZNickq:
Looks good, thanks ;)

PR #233, opened 2012-06-04
Spawn eggs now spawn mobs when used

zhuowei:
This allows a player to use spawn eggs. The code is based on the painting placement code, and will allow unlimited uses if the player is in creative mode.
Signed-off-by: Zhuowei Zhang < zhuowei zhang @ ya hoo.com >


ZNickq:
Thanks, looks good!

PR #234, opened 2012-06-05
Fixed issues with casting.

Pamelloes:
ZIdane had some... difficulties with understanding how casting objects works. I remedied the situation. :P

Signed-off-by: Pamelloes joshua314@sbcglobal.net


Zidane:
@Pamelloes 

Not at all...you actually decided to completely go the WRONG WAY and store what should only be a byte into a full blow integer.

My issue is that I completely forgot we didn't have a getByte and decided to improvise without casting the default in BiomeConfiguration to Byte.

Pamelloes:
@NinjaZidane I didn't make any changes, I just made your code work. Either way, I think @bergerkiller handled it.

PR #237, opened 2012-06-09
Refactored GameMode to be a class.

Pamelloes:
GameMode is now configurable in Vanilla's config. Plugins can now register custom GameModes. In the future, the customizability behind GameMode will be expanded.

On a side note, PLEASE read over my code and make sure I did everything right! I made a lot of changes and there is now way I sufficiently tested each one. If any changes are needed, I will be happy to make them.

### DIFF
--- a/src/main/java/org/spout/vanilla/VanillaPlugin.java
+++ b/src/main/java/org/spout/vanilla/VanillaPlugin.java
@@ -147,7 +147,7 @@ public class VanillaPlugin extends CommonPlugin {
 		if (WorldConfiguration.NORMAL_LOAD.getBoolean()) {
 			NormalGenerator normGen = new NormalGenerator();
 			World normal = game.loadWorld(WorldConfiguration.NORMAL_NAME.getString(), normGen);
-			normal.getDataMap().put(VanillaData.GAMEMODE, GameMode.valueOf(WorldConfiguration.NORMAL_GAMEMODE.getString().toUpperCase()));
+			normal.getDataMap().put(VanillaData.GAMEMODE, GameMode.getByName(WorldConfiguration.NORMAL_GAMEMODE.getString().toUpperCase()));
 			normal.getDataMap().put(VanillaData.DIFFICULTY, Difficulty.valueOf(WorldConfiguration.NORMAL_DIFFICULTY.getString().toUpperCase()));
 			normal.getDataMap().put(VanillaData.DIMENSION, Dimension.valueOf(WorldConfiguration.NORMAL_SKY_TYPE.getString().toUpperCase()));
 			//Grab safe spawn if newly created world.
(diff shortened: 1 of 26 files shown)


Pamelloes:
I've been told this needs rebasing, but I have no idea how to do that.... Last time, the instructions I received from @W1ndwaker were....... unsuccessful.

ZNickq:
try git pull --rebase origin

Pamelloes:
Thanks @ZNickq should be pull-able now.

src/main/java/org/spout/vanilla/data/GameMode.java
```diff
@@ -26,31 +26,142 @@
...
 	}
+	
+	/**
+	 * Gets the GameMode with the given name.
+	 */
+	public static GameMode getByName(String name) {
+		return nameMap.get(name.toUpperCase());
+	}
+	
+	public static enum VanillaGameMode {
```
windy1, line 142: What is the point of this enum no that we have a Creative and Survival GameMode classes

windy1:
I'de imagine that the GameMode itself should actually have an onTick instead of checking the player's gamemode in VanillaPlayer; that way we can just pass `gameMode.onTick(float dt)` in VanillaPlayer#onTick instead of all those "has" checks.

src/main/java/org/spout/vanilla/data/GameMode.java
Pamelloes, line 142: The class is used for creating custom game modes, this enum defines the minecraft ones which are used as a backing for sending to the client. So, when I create a custom game mode, lets just say EndGameMode, I need to specify a VanillaGameMode (Survival). That way, when I assign a player to EndGameMode which doesn't exist in Vanilla minecraft, the client thinks its in survival. VanillaGameMode is an enum because the Vanilla game modes can't be changed, whereas GameMode is a class to allow for extensibility.

Pamelloes:
@W1ndwaker that's a good idea, it will be a major refactor, though. I will get around to it sometime soon.

Pamelloes:
@W1ndwaker On second thought it is a very big refactor that is beyond the scope of this PR. I will make another one in the future that expands the capabilities of GameMode, the onTick method will be put there.

windy1:
Okay sounds good. I'll close this one then.

Pamelloes:
@W1ndwaker hrm? I meant I was going to make a second PR after this one got pulled.... :(

windy1:
Why would I pull something that's not complete yet? I'de rather you make it one pull.

Pamelloes:
@W1ndwaker, its complete, there's just room for improvement. (And it would save me the work of rebasing again :P)

windy1:
@Pamelloes It's not a big deal, why would I pull it if the system we have now is not broken, and a lot of this system will most likely be replaced anyways. (And it's not that difficult :p)

ZNickq:
GameMode was actually a class some time ago, and it allowed all that stuff (custom gamemodes), but i changed it to an enum later, not sure why :/

windy1:
It was me actually Znickq the implementation was different then this as well. I agree that this would be better but still needs improvement and is not quite worth pulling yet.

src/main/java/org/spout/vanilla/VanillaPlugin.java
```diff
@@ -186,7 +186,7 @@ private void setupWorlds() {
 		if (WorldConfiguration.END_LOAD.getBoolean()) {
 			TheEndGenerator endGen = new TheEndGenerator();
 			World end = game.loadWorld(WorldConfiguration.END_NAME.getString(), endGen);
-			end.getDataMap().put(VanillaData.GAMEMODE, GameMode.valueOf(WorldConfiguration.END_GAMEMODE.getString().toUpperCase()));
+			end.getDataMap().put(VanillaData.GAMEMODE, GameMode.getByName(WorldConfiguration.END_GAMEMODE.getString().toUpperCase()));
```
windy1, line 189: What's the point of adding a getByName method?

ZNickq, line 189: Well, it's used right here...o.O

windwaker, line 189: Guys,

Ye are including the wrong 'Windwaker' on this thread. I think ye are
looking for 'W_1_ndwaker'

On Tue, Jun 12, 2012 at 3:58 PM, ZNickq <
reply@reply.github.com

windy1, line 189: Sorry about that @windwaker 

And @znickq I was asking why use this instead of GameMode#valueOf

Pamelloes, line 189: @W1ndwaker valueOf is an enum method and doesn't exist in the GameMode class.

windy1, line 189: I know that

EDIT: Oh I see it's referencing the class not the enum, derp.

PR #238, opened 2012-06-13
More Event API

IronCrystal:
I added the following files:

PlayerEnchantItemEvent.java
EntityGameModeChangeEvent.java
EntityShootProjectileEvent.java
EntityVehicleEvent.java


src/main/java/org/spout/vanilla/event/entity/EntityGameModeChangeEvent.java
```diff
@@ -0,0 +1,78 @@
...
+ * including the MIT license.
+ */
+package org.spout.vanilla.event.entity;
+
+import org.spout.api.event.HandlerList;
+import org.spout.api.event.Cancellable;
+import org.spout.api.event.entity.EntityEvent;
+import org.spout.api.entity.Entity;
+
+public class EntityGameModeChangeEvent extends EntityEvent implements Cancellable {
```
windy1, line 34: GameMode's will most likely be undergoing a re factor shortly, so I'll hold of on this for now.

src/main/java/org/spout/vanilla/event/player/PlayerEnchantItemEvent.java
```diff
@@ -0,0 +1,149 @@
...
+	@Override
+	public HandlerList getHandlers() {
+		return handlers;
+	}
+
+	public static HandlerList getHandlerList() {
+		return handlers;
+	}
+	
+	public enum Enchantment {
```
windy1, line 122: We want to be able to obtain the Enchantment object, not an enum.

IronCrystal:
So, what do I do?  Do I do another PR with just the three files (not the game mode) and fix the Enchantment object?

windy1:
Sure, that's probably the best course of action since you accidentally included one of @DDoSqc's commits

PR #239, opened 2012-06-14
Added More Event API for Entity and Player

IronCrystal:
I added the following Event API

PlayerEnchantItemAPI.java
EntityShootProjectileEvent.java
EntityVehicleEvent.java

I fixed the Enchantment Object.

### DIFF
--- a/src/main/java/org/spout/vanilla/event/entity/EntityShootProjectileEvent.java
+++ b/src/main/java/org/spout/vanilla/event/entity/EntityShootProjectileEvent.java
@@ -1,0 +1,68 @@
+/*
+ * This file is part of Vanilla.
+ *
+ * Copyright (c) 2011-2012, VanillaDev <http://www.spout.org/>
+ * Vanilla is licensed under the SpoutDev License Version 1.
+ *
+ * Vanilla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In addition, 180 days after any changes are published, you can use the
+ * software, incorporating those changes, under the terms of the MIT license,
+ * as described in the SpoutDev License Version 1.
+ *
+ * Vanilla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License,
+ * the MIT license and the SpoutDev License Version 1 along with this program.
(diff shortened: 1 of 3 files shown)


src/main/java/org/spout/vanilla/event/entity/EntityVehicleEvent.java
```diff
@@ -0,0 +1,144 @@
...
+import org.spout.api.event.entity.EntityEvent;
+
+public class EntityVehicleEvent extends EntityEvent implements Cancellable {
+
+	private static HandlerList handlers = new HandlerList();
+	private Controller ridden;
+	private Controller rider;
+	private VehicleReason reason;
+
+	public EntityVehicleEvent(Entity e, Controller ridden, Controller rider, VehicleReason reason) {
```
Zidane, line 43: Good changes all around through here...except when we talked about it I plum forgot about the forced passing of an entity to the super.

In some ways, I think the core of events need to be heavily looked at. We are aiming for a "Entity component" model and this goes against it

Thoughts @W1ndwaker ?

IronCrystal, line 43: Yeah I was a bit confused because it extends EntityEvent, so it has to super entity.

src/main/java/org/spout/vanilla/event/entity/EntityShootProjectileEvent.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package org.spout.vanilla.event.entity;
+
+import org.spout.api.entity.Entity;
+import org.spout.api.event.HandlerList;
+import org.spout.api.event.Cancellable;
+import org.spout.api.event.entity.EntityEvent;
+import org.spout.vanilla.controller.object.Projectile;
+
+public class EntityShootProjectileEvent extends EntityEvent implements Cancellable {
```
windy1, line 35: EDIT: Scratch that, i suppose this makes sense actually.

src/main/java/org/spout/vanilla/event/entity/EntityVehicleEvent.java
windy1, line 43: When is this supposed to be called?

IronCrystal, line 43: Zidane wanted like one general event, so I guess this is called any time a vehicle does anything.  Ie creation, collision (which isn't implemented yet) exit and enter, etc

Afforess, line 43: VehicleStateChangeEvent?

Zidane, line 43: Agreed with @Afforess

windy1:
Do you intend on making the proposed changes @IronCrystal or should this be closed?

IronCrystal:
You can close this

PR #241, opened 2012-06-19
I added online mode

ezruneko:


ZNickq:
Looks good, but you completely changed the formatting of one of the files :P

Raphfrk:
The first file should have minimum changes.

However, we already supported online mode.

```
public static final ConfigurationHolder ONLINE_MODE = new ConfigurationHolder(true, "general", "online-mode");
```

ezruneko:
I use eclipse on windows so he changes \r\n :(

"However, we already supported online mode.". How? i dont see it on any configuration file.

Raphfrk:
The parameter is the one I quoted.

Also, online mode is handled by the packets accepting the handshake:

https://github.com/VanillaDev/Vanilla/blob/master/src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapHandshakeMessageHandler.java

Raphfrk:
One thing it doesn't support is offline and encryption mode.  However, I am not sure what that would achieve.

It would prevent the server from verifying that a user is who they claim to be.  A trusting server like that isn't going to benefit from encrypting the connection much.

stephenwilliams:
Raphfrk you can delete and edit posts

Raphfrk:
Thanks, not sure what is happening here, looks like my browser won't auto-update, when I comment.

Zidane:
Already in

PR #246, opened 2012-06-28
Update Recipes List

SonarBeserk:
Recipes list for the Vanilla plugin.

### DIFF
--- a/src/main/resources/recipes.yml
+++ b/src/main/resources/recipes.yml
@@ -2,319 +2,15 @@ recipes:
 
   Plank:
     type: Shapeless
-    result: PLANK,1
+    result: PLANK,4
     ingredients: [LOG]
-   
-  Block_GOLD:
-    type: Shaped
-    result: GOLDBLOCK,1
+    
+  Stick:
+    type: Shapeless
+    result: STICK,4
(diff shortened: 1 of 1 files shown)


kitskub:
http://www.minecraftwiki.net/wiki/Crafting
That is the complete recipe list.
Also, the commits need to be squashed.

SonarBeserk:
latest commit is the current one.
@kitskub i need you to confirm the name for the slabs. and most likely stairs so i can complete it. What Result do i need to return for those?

kitskub:
I just looked at slabs, but they are in the form of type slab, (i.e. sandstone slab). You can check it in material/block/misc. And again, these need to be a single commit. It is showing three different ones.

Salynne:
There was another pull that was more correct and already done so closing this. Thanks though.

SonarBeserk:
my question is if it was done why is the recipes so out of date?

Afforess:
The recipes seem up to date to me: https://github.com/VanillaDev/Vanilla/blob/master/src/main/resources/recipes.yml

SonarBeserk:
they wernt before. they are now. so it is finished.

PR #247, opened 2012-06-29
Cleanup warnings

tips48:
I cleaned up most of the warnings in all the Spout projects. It was mainly imports, but there were some issues with not closing resources and type casting.
Enjoy!


Zidane:
Thanks for the PR but I will do this when I run the formatter today :)

PR #248, opened 2012-07-02
Added all recipes as of Minecraft 1.2.5

Owexz:
Signed-off-by: Owexz Owexz@Owexz.net


kitskub:
Unfortunately, I believe this is not correct. The parser gets the material not by the variable name, but by the registered name. For example, white wool would not be "WHITE_WOOL" but "White Wool."

Afforess:
For future readers, I tweaked the material parser to match names a bit more greedily, so @kitskub concerns were valid, but with my changes "Iron Block" and "IRON_BLOCK" both parse to the same, correct material.

PR #255, opened 2012-07-12
Universal Plug and Play by using cling libraries

Darsstar:
InetAddress.getLocalHost() is used because mapping the port to 0.0.0.0 did not seem to work. Which would mean that UPNP would not work out of the box since 0.0.0.0:25565 is the default spout address.

Friends were able to join when the upnp setting was true.

I'm not sure if putting everything in Vanilla is the best idea, maybe cling should be shaded to SpoutAPI, or maybe to Spout with some method added to the API.

And last but not least, a related Spout issue: http://issues.spout.org/browse/SPOUT-71

Related:
https://github.com/SpoutDev/Spout/pull/1937
https://github.com/SpoutDev/SpoutAPI/pull/289

### DIFF
--- a/src/main/java/org/spout/vanilla/VanillaPlugin.java
+++ b/src/main/java/org/spout/vanilla/VanillaPlugin.java
@@ -85,6 +85,7 @@ public class VanillaPlugin extends CommonPlugin {
 	private Engine engine;
 	private VanillaConfiguration config;
 	private static VanillaPlugin instance;
+	private int port = 25565;
 
 	@Override
 	public void onDisable() {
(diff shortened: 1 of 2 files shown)


Afforess:
This should really be a part of the engine, not Vanilla! Cool stuff though.

lukespragg:
I agree with Afforess. This is something for the engine, not Vanilla specific.

Zidane:
I talked with @Darsstar earlier and I think he is going to put it into the API/Engine. Agreed definitely with the above...love this feature.

Spouty:
**[AUTO-GENERATED COMMENT]**

Please make sure this pull request follows the [pull request guidelines](http://spout.in/prguide) in order to ensure your pull request is accepted.

This Pull Request had errors on compile! See the [log](http://pr.spouty.org/files/?file=3248405133641172/log.txt) for more details

Darsstar:
The Spout and SpoutAPI will be up as soon as I manage to correct the line-endings...

Darsstar:
They are all done, for links scroll to the top.

dredhorse:
NIce!

src/main/java/org/spout/vanilla/VanillaPlugin.java
```diff
@@ -105,6 +106,14 @@ public void onEnable() {
 		} catch (ConfigurationException e) {
 			getLogger().log(Level.WARNING, "Error loading Vanilla configuration: ", e);
 		}
+
+		//Universal Plug and Play
+		if (engine.getPlatform() == Platform.SERVER || engine.getPlatform() == Platform.PROXY) {
```
Zidane, line 111: I am a bit loopy from medicine right now so forgive me if this is obvious but...why would Spout run plugins in proxy-mode?

zhuowei, line 111: To provide protocols for the proxy. e.g. to proxy Minecraft, you need Vanilla.

Darsstar, line 111: I copied that if statement from the onLoad method. I figured anytime you want to be able to bind an address you will also want to map a port for it.

PR #259, opened 2012-07-18
Generic Logic Class for Creatures dropping items tick-based.

timmhirsens:
Generic Logic Class for Creatures dropping items tick-based.

Also made the Chicken use the new Logic-Class.

Signed-off-by: fr1zle timmhirsens@gmail.com


kitskub:
This needs to be one commit.

timmhirsens:
Yup, had some issues with line endings. Should be fixed now?

PR #261, opened 2012-07-23
Fixed Creature health

greatman:
All creature healths was at 1 because of the super.onAttached() being called after and also because of setMaxHealth being set after setting the health.

This pull fix this issue.


greatman:
Also forgot, I set the Ocelot health to 10 directly since it's only that. It's not the same behavious as a Wolf.

PR #263, opened 2012-07-24
onTick takes care of the kill and other fixes

greatman:
onTick() takes care of the kill.

Fixed Minecart regen HP. May need some tweak but looks like Vanilla to my eye. Also the Minecart now drop a minecart.

Added a DeathAnimation flag. All Living entities got that flag so it flips on his side while dying.


windy1:
Wouldnt it have made more sense to make death animation true by default?

greatman:
By my eye I was thinking there was more entities that don't have a death animation. That's why I set it to false.

PR #266, opened 2012-07-25
1.3 Login auth looks working. Need a working MapData protocol to really ...

greatman:
...test.

Bumped the Minecraft protocol ID

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapEncryptionKeyResponseMessageHandler.java
+++ b/src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapEncryptionKeyResponseMessageHandler.java
@@ -31,19 +31,15 @@ import java.security.MessageDigest;
 
 import org.bouncycastle.crypto.AsymmetricBlockCipher;
 import org.bouncycastle.crypto.AsymmetricCipherKeyPair;
-import org.bouncycastle.crypto.BufferedBlockCipher;
-import org.bouncycastle.crypto.CipherParameters;
-import org.bouncycastle.crypto.params.KeyParameter;
-import org.bouncycastle.crypto.params.ParametersWithIV;
 
 import org.spout.api.Spout;
+import org.spout.api.event.Event;
+import org.spout.api.event.player.PlayerConnectEvent;
 import org.spout.api.player.Player;
 import org.spout.api.protocol.MessageHandler;
 import org.spout.api.protocol.Session;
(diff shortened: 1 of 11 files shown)


Raphfrk:
Geez, I go to sleep and someone else does it :).

src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapEncryptionKeyResponseMessageHandler.java
```diff
@@ -68,47 +64,45 @@ public void handleServer(final Session session, final Player player, final Encry
...
+			final byte[] savedValidateToken = (byte[]) session.getDataMap().get("verifytoken");
 
-						EncryptionChannelProcessor fromClientProcessor = new EncryptionChannelProcessor(fromClientCipher, 32);
-						EncryptionChannelProcessor toClientProcessor = new EncryptionChannelProcessor(toClientCipher, 32);
+			boolean equals = true;
+			for (int i = 0; i < validateToken.length && equals; i++) {
+				if (validateToken[i] != savedValidateToken[i]) {
+					equals = false;
+				}
+			}
```
Raphfrk, line 83: Why not just drop the equals variable and simply replace equals = false with

```
kickInvalidUser(session);
return
```

src/main/java/org/spout/vanilla/protocol/bootstrap/handler/BootstrapEncryptionKeyResponseMessageHandler.java
```diff
@@ -68,47 +64,45 @@ public void handleServer(final Session session, final Player player, final Encry
...
 
-					session.disconnect("Encryption not supported yet");
-				}
-			};
+			String handshakeUsername = session.getDataMap().get(VanillaProtocol.HANDSHAKE_USERNAME);
+			final String finalName = handshakeUsername.split(";")[0];
+			Thread loginAuth = new Thread(new LoginAuth(session, finalName, new PlayerConnectRunnable(session, finalName)));
+			loginAuth.start();
+			while (loginAuth.isAlive()) {
+			}
```
Raphfrk, line 99: The login auth thread is supposed to send the response, not the current thread.

```
while (thread.isAlive()) {}
```

spins the current thread until the other one is done.  If you have to do it that way, then you should use loginAuth.join(), which means wait until that other thread is dead.

However, waiting on this thread defeats the whole purpose of doing the login auth in another thread.

PR #268, opened 2012-07-25
Open Auth Server

ezruneko:


Raphfrk:
Hmm, so allows more 1 authentication server?

A new handler would be required based on name.

Btw, some description for a PR is helpful :).

ezruneko:
yep, allow more than an auth server.

LoginAuth is the template to new implements new Auth servers.
To register by name use LoginAuthRegistry.register(name, newLoginAuth.class);
And to swith betweens Auth server for the user is easy only add on vanilla config:

auth-server: myloginserver

Raphfrk:
So, this would allow authing to be handled by plugins?

This is a fine line really.  The current policy is that Vanilla will use the official server for authing.

However, maybe that could just be for default.

ezruneko:
yes, it can be handled by plugins.

olliestanley:
@Raphfrk

Online
Afforess
Administrator
Spout Team
Achievements:
8
Ratings Received:
+1,082 / 106

Kohle said: ↑
Yes, a Mojang account is still needed for Vanilla.

Initially yes, although I don't think this will always be the case. I personally don't see a need for it, right now we do it just as a way to verify players are who they say they are, but ideally we will transition to a Spout based account system in the future.

From http://forums.spout.org/threads/curse.4211/page-3

ezruneko:
some alternatives appear to authlike https://github.com/KayoticSully/Open-Mineshafter-Squared/zipball/master or http://mineshafter.appspot.com/

Raphfrk:
@DziNeIT , yeah that is new :p.

olliestanley:
True, I was just putting it here for your information :)

ezruneko:
No accepted or rejected this pull?

Raphfrk:
We haven't really had any more discussion on how we are handling auth.

ezruneko:
ok, np

ZNickq:
I'll just go ahead and close this, when we decide on handling auth, we'll see if this can be implemented ;)

PR #269, opened 2012-07-25
Emerald addition & Protocol update

greatman:
Added Emerald.
Added SlotData field type for the Protocol (Untested)
Updated EntityEquipment


Raphfrk:
You need to use -s to add the signoff message

greatman:
Oh right forgot about it. Added ;)

PR #270, opened 2012-07-25
Implemented basic pathfinding.

timmhirsens:
Basic finding for creatures using the A*-Algorithm.


timmhirsens:
Note: this is a very basic implementation. It might not be the best, but it should work for randomly walking passive creatures.

The reason why I'm already openening this PR is that will be on holidays for the next 1 1/2 weeks. So if anyone wants to work on that he should have good starting point.

PR #294, opened 2012-08-17
Spawn protection for placing blocks. Fixes VANILLA-159

greatman:
Signed-off-by: William Lebel lebelwilliam@gmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/handler/PlayerBlockPlacementMessageHandler.java
+++ b/src/main/java/org/spout/vanilla/protocol/handler/PlayerBlockPlacementMessageHandler.java
@@ -26,10 +26,15 @@
  */
 package org.spout.vanilla.protocol.handler;
 
+import java.util.Collection;
+
+import org.spout.api.Spout;
+import org.spout.api.chat.style.ChatStyle;
 import org.spout.api.entity.Player;
 import org.spout.api.event.EventManager;
 import org.spout.api.event.player.PlayerInteractEvent;
 import org.spout.api.event.player.PlayerInteractEvent.Action;
+import org.spout.api.geo.Protection;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/protocol/handler/PlayerBlockPlacementMessageHandler.java
```diff
@@ -193,7 +199,17 @@ public void handleServer(Session session, PlayerBlockPlacementMessage message) {
...
 					}
 				}
-
+				
+				//Check if the player can place the block.
+				Collection<Protection> protections = Spout.getEngine().getServiceManager().getRegistration(ProtectionService.class).getProvider().getAllProtections(alterBlock.getPosition());
+				for (Protection p : protections) {
+					if (p.contains(alterBlock.getPosition()) && !((VanillaPlayerController)player.getController()).isOp()) {
+						undoPlacement(player, clickedBlock, alterBlock);
+						player.sendMessage(ChatStyle.DARK_RED, "This area is a protected spawn point!");
```
ZNickq, line 208: Doesn't this check all the protections, not just the spawn point?

greatman, line 208: Well I copied the same code as in the PlayerDiggingMsg.

ZNickq, line 208: @NinjaZidane wrote that code i believe...

Zidane, line 208: It needs to be re-written with API in mind.

PR #300, opened 2012-08-28
Added customizable outdated client / server messages

dredhorse:
This is a often requested feature for MessageChanger, which isn't possible, but we can do it directly in Vanilla.

What would be cool that to a later point in time the Spoutclient allows clickable links.

Signed-off-by: dredhorse dredhorse@breiden.net


windy1:
Awesome, just one thing: I would prefer if the nodes were under the general category.

dredhorse:
done...

PR #301, opened 2012-08-29
change the encoding of chatFlags to byte from int

aegonthemad:
Signed-off-by: AegonTheMad aegonthemad@gmail.com


windy1:
Nice catch ;)

PR #303, opened 2012-08-31
Components

dredhorse:
added missing creatures and optimized imports


dredhorse:
argh... where are the other commits coming from?

Zidane:
Gonna close this for you @dredhorse

dredhorse:
still wonder how the correct way to do this is... because next time it will not be so easy

PR #312, opened 2012-09-09
Fixed brewing stand output index check

kitskub:
There are only three outputs, so it should only be 0, 1, or 2


windy1:
There are three outputs... 0, 1, and 2. :p

kitskub:
Yes... think again. It has to be greater than or equal to 0 and less than 3...

windy1:
Oh herp I read the diff incorrectly.. my mistake.

kitskub:
Lol

PR #318, opened 2012-09-15
Added /plugins-tofile [filename].

luca-moser:
This command outputs a file having informations about all loaded plugins. 
The report is stored in plugins/Vanilla/pluginreports. 
The commadsource can define the filename of the report but if no filename 
was passed up on execution, then the current date and time is taken as the filename for the report.

The informations stored into the report are: name, version and authors(if info available).

An example outputfile looks like this:
http://www.hastebin.com/metukahupe.1c

Signed-off-by: Luca Moser moser.luca@gmail.com


DDoS:
Only one commit please.
And the formatting is off (spaces are missing).

luca-moser:
Will create a new PR.

PR #319, opened 2012-09-15
Added /plugins-tofile [filename].

luca-moser:
This command outputs a file having informations about all loaded plugins.
The report is stored in plugins/Vanilla/pluginreports.
The commadsource can define the filename of the report but if no filename
was passed up on execution, then the current date and time 
is taken as the filename for the report.

The informations stored into the report are: name, version and authors(if info available).

An example outputfile looks like this:
http://www.hastebin.com/metukahupe.1c

Signed-off-by: Luca Moser moser.luca@gmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/command/TestCommands.java
+++ b/src/main/java/org/spout/vanilla/command/TestCommands.java
@@ -26,13 +26,6 @@
  */
 package org.spout.vanilla.command;
 
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.Date;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/command/TestCommands.java
```diff
@@ -262,4 +270,111 @@ public void debug(CommandContext args, CommandSource source) throws CommandExcep
...
+        }
+
+        source.sendMessage("Plugins-report successfully created! " + linesep + "Stored in: " + standpath);
+    }
+
+    /**
+     * Replaces chars which are not allowed in filenames on windows with "-".
+     * Well atleast some
+     */
+    public String replaceInvalidCharsWin(String s) {
```
DDoS: Is it really necessary for this method to be public?

luca-moser: No actually not, haha. Will change this.

luca-moser:
Another close.

PR #324, opened 2012-10-02
Fix World can get from the player VANILLA-200

ghost:


Zidane:
I'd like to pull this but I am afraid you didn't sign it off. We require this as it tells us you agree to our license.

Read the guidelines here: https://github.com/VanillaDev/Vanilla/blob/master/README.md (at the bottom)

ghost:
Done, i think.

Zidane:
I actually pushed it in my changes as I had already made the change locally a bit back and never commited it. I still gave you credit in the push I did so thanks.

Will go ahead and close this.

PR #329, opened 2012-10-08
Added EntityProtocolID enum to name the IDs associated with specific

Sleaker:
Vanilla mobs.
Added Bat.
Changed all Entities to use EntityProtocolID instead of a direct int value for code legibility.


windy1:
Might as well include the object entities while you're at it to, eh? ;)

And I think I would prefer it if the protocols could take the enum as a param instead of having to do `enum.getId()` everywhere.

Sleaker:
@W1ndwaker - yah been talking with Zidane on this, I'm going through everything right now on the 1.4 Entities and adding them in.

The issue with the Enum param is that it would need to be built into the Parameter class. can be done, but not necessarily something needed?

windy1:
Yeah I'm aware @Sleaker, just thought it would be neater to pass in an enum rather then an ID.

Raphfrk:
Hmm, are we sure we want to go down the enum route.  There was some concerns that it prevents extension, since they are fixed.

However, extensibility isn't as critical for Vanilla as for the core server.

Sleaker:
@Raphfrk - the enum in this situation is only for eliminating Magic Numbers, it provides pretty much no other use here.  That's also why I don't necessarily think the Protocols need to accept the enum.

src/main/java/org/spout/vanilla/component/living/hostile/Zombie.java
```diff
@@ -29,15 +29,34 @@
...
+	public boolean wasVillager() {
+		return villager;
+	}
+	
+	/**
+	 * Sets if this is a villager zombie.
+	 * @param value
+	 */
+	public void setWasVillager(boolean value) {
+		villager = value;
```
Zidane: Store this in the datatable

Add a DefaultedKey in VanillaData.

src/main/java/org/spout/vanilla/component/living/passive/Villager.java
```diff
@@ -65,4 +65,12 @@ public CraftingInventory getInventory() {
 		// TODO Auto-generated method stub
 		return null;
 	}
+	
+	/**
+	 * The integer value associated with this villager type.
+	 * @return int 
+	 */
+	public int getVillagerTypeID() {
+		return 5; // Generic Villager ID.
```
Zidane, line 74: Store the ID in the datatable else this won't be persisted.

src/main/java/org/spout/vanilla/data/EntityProtocolID.java
```diff
@@ -0,0 +1,42 @@
+package org.spout.vanilla.data;
```
Zidane, line 27: Need a header here :p

Sleaker, line 27: no it don't! shhh.

src/main/java/org/spout/vanilla/component/living/passive/Villager.java
windy1, line 74: ummm, it's a constant, every time a new Villager is created `getVillagerTypeId()` will always return 5...

src/main/java/org/spout/vanilla/component/living/hostile/Skeleton.java
```diff
@@ -38,6 +39,6 @@
 	@Override
 	public void onAttached() {
 		super.onAttached();
-		getHolder().getNetwork().setEntityProtocol(VanillaPlugin.VANILLA_PROTOCOL_ID, new BasicMobEntityProtocol(51));
+		getHolder().getNetwork().setEntityProtocol(VanillaPlugin.VANILLA_PROTOCOL_ID, new SkeletonEntityProtocol(EntityProtocolID.SKELETON.getId()));
```
windy1, line 42: Must we have specific protocols for entities with no metadata? There's really no need. http://wiki.vg/Entities

src/main/java/org/spout/vanilla/data/EntityProtocolID.java
windy1, line 27: Also, we might just want to call it EntityType or something like that, since it's not the actual id itself. Doesn't `EntityType.BAT.getId()` make more sense then `EntityProtocolID.BAT.getId();`?

EDIT: Also, underscores for multiple words please :P (`CAVE_SPIDER`, `ENDER_DRAGON`, `IRON_GOLEM`, etc)

src/main/java/org/spout/vanilla/protocol/entity/living/SkeletonEntityProtocol.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.spout.api.entity.Entity;
+import org.spout.api.util.Parameter;
+
+import org.spout.vanilla.component.living.hostile.WitherSkeleton;
+import org.spout.vanilla.data.EntityProtocolID;
+import org.spout.vanilla.protocol.entity.BasicMobEntityProtocol;
+
+public class SkeletonEntityProtocol extends BasicMobEntityProtocol {
+	public final static int TYPE_INDEX = 13; // The MC metadata index for determining the type of the skeleton
```
windy1, line 39: Ah, does Skeleton have metadata in 1.4? If so please disregard my previous comment.

windy1:
@Sleaker It was just a design thing, I was proposing to add a new constructor for the enum, not remove the existing one. Also, if it were me, I would have an

``` java
interface EntityType {
    int getId();
}
```

then

``` java
enum CreatureType implements EntityType {
    // blah, blah, blah

    final int id;

    CreatureType(int id) {
        this.id = id;
    }

    @Override
    int getId() {
        return id; 
    }
}
```

then we could do all sorts of fun OO stuff! 
and the same for ObjectType. But again just semantics here :p

src/main/java/org/spout/vanilla/component/living/hostile/Skeleton.java
Sleaker, line 42: There is, due to differences in Ai for future-proofing. this was discussed a bit between Zidane and I.

src/main/java/org/spout/vanilla/protocol/entity/living/SkeletonEntityProtocol.java
Sleaker, line 39: yah it does now. Wither Skeleton = Skeleton with metadata == 1

src/main/java/org/spout/vanilla/component/living/passive/Villager.java
Sleaker, line 74: Yah miscommunication there. He didn't realize I had made sub-classes for each villager type. Reason again different AI on each one and they all interact completely different.

Sleaker:
Closing and re-opening with squashed files.

PR #332, opened 2012-10-11
Prevents Nether Portals from being created if they would touch another

megagamer410:
nether portal. Consistent with (half) of default minecraft functionality, and prevents a rendering error where having 2 nether portals facing the same direction touching each other would cause the portals to become invisible.

Signed-off-by: megagamer410 megagamer410@gmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
+++ b/src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
@@ -33,6 +33,7 @@ import org.spout.api.material.block.BlockFaces;
 import org.spout.vanilla.data.MoveReaction;
 import org.spout.vanilla.material.VanillaMaterials;
 import org.spout.vanilla.material.block.Portal;
+import org.spout.vanilla.world.generator.theend.TheEndGenerator;
 
 public class NetherPortal extends Portal {
 	public NetherPortal(String name, int id) {
@@ -56,6 +57,10 @@ public class NetherPortal extends Portal {
 	 * @return True if it was successful, False if not
 	 */
 	public boolean createPortal(Block bottomBlock) {
+		//  Check if the portal is being created in the End.
+		if (bottomBlock.getWorld().getGenerator() instanceof TheEndGenerator) {
+			return false;
+		}
 		Block above = bottomBlock.translate(BlockFace.TOP);
 		// Find out what direction to create the portal to
 		BlockFace direction = null;
@@ -98,6 +103,21 @@ public class NetherPortal extends Portal {
 		}
 		if (!corner2.isMaterial(VanillaMaterials.OBSIDIAN)) {
 			return false;
+		}
+		// Validate that there are no portals touching the proposed one.
+		corner1 = above;
+		corner2 = above.translate(direction);
+		for (int i = 0; i < 3; i++) {
+			// Check if there is another nether portal touching the proposed one
+			for (BlockFace side : BlockFaces.NESW) {
+				// If so, don't build the current one, and extinguish the fire.
+				if (corner1.translate(side).isMaterial(VanillaMaterials.PORTAL) || corner2.translate(side).isMaterial(VanillaMaterials.PORTAL)) {
+					above.setMaterial(VanillaMaterials.AIR);
+					return false;
+				}
+			}
+			corner1 = corner1.translate(BlockFace.TOP);
+			corner2 = corner2.translate(BlockFace.TOP);
 		}
 		// Validated, time to create the portal
 		corner1 = above;
--- a/src/main/java/org/spout/vanilla/material/item/tool/FlintAndSteel.java
+++ b/src/main/java/org/spout/vanilla/material/item/tool/FlintAndSteel.java
@@ -55,11 +55,6 @@ public class FlintAndSteel extends InteractTool {
 				// Default fire creation
 				Block target = block.translate(clickedface);
 
-				// Handle the creation of portals
-				if (VanillaMaterials.PORTAL.createPortal(target.translate(BlockFace.BOTTOM))) {
-					return;
-				}
-
 				// Default fire placement
 				clickedface = clickedface.getOpposite();
 				if (VanillaMaterials.FIRE.canPlace(target, (short) 0)) {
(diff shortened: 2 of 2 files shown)


src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
```diff
@@ -99,10 +99,23 @@ public boolean createPortal(Block bottomBlock) {
 		if (!corner2.isMaterial(VanillaMaterials.OBSIDIAN)) {
 			return false;
 		}
-		// Validated, time to create the portal
+		// Validate that there are no portals touching the proposed one.
+		// NOTE: this matches the default minecraft functionality only on portals facing North-South
```
DDoS: Yeah, not sure about the half compatibility thing. It would be better to do it now entirely, instead of doing it partially, forgetting about it, and then finally fixing it later...

megagamer410: The problem with that is that the default minecraft implementation isn't particularly logical. If the portals face North / South then it doesn't create the new one, but if it's East / West then it creates the new one, unless it's being created between 2 existent portals, at which case it destroys the west most one and fails to build the proposed one. I opted to only commit the North / South implementation because it makes the most sense, but I'll start working on the East / West compatibility.

Raphfrk: I don't think we need an exact match if it makes no sense.  @zidane can comment.

Does this also include creating portals in other worlds?

megagamer410: Portal creation is identical in the Overworld and the Nether but fails in the End, so I updated the commit to reflect that.

windy1: Are nether portals supposed to work in the end or is a bug in Minecraft?

megagamer410: Nether portals are not supposed to work in the end.

src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
```diff
@@ -56,6 +57,10 @@ public MoveReaction getMoveReaction(Block block) {
 	 * @return True if it was successful, False if not
 	 */
 	public boolean createPortal(Block bottomBlock) {
+		//  Check if the portal is being created in the End.
+		if (bottomBlock.getBiomeType().getClass().equals(EndStoneBiome.class)) {
```
windy1: I think

``` java
if (world.getGenerator() instanceof TheEndGenerator) {
    // do stuff
```

would be more logical here.

windy1:
And I presume you've tested this as well? :p

src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
```diff
@@ -56,6 +57,10 @@ public MoveReaction getMoveReaction(Block block) {
 	 * @return True if it was successful, False if not
 	 */
 	public boolean createPortal(Block bottomBlock) {
+		//  Check if the portal is being created in the End.
+		if (bottomBlock.getWorld().getGenerator() instanceof TheEndGenerator) {
+			return false;
+		}
```
megagamer410, line 63: This prevents Nether Portals from being created in the End, but leaves the fire. (Same as Minecraft)

src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
```diff
@@ -99,6 +104,21 @@ public boolean createPortal(Block bottomBlock) {
...
+			// Check if there is another nether portal touching the proposed one
+			for (BlockFace side : BlockFaces.NESW) {
+				// If so, don't build the current one, and extinguish the fire.
+				if (corner1.translate(side).getMaterial().equals(VanillaMaterials.PORTAL) || corner2.translate(side).getMaterial().equals(VanillaMaterials.PORTAL)) {
+					above.setMaterial(VanillaMaterials.AIR);
+					return false;
+				}
+			}
+			corner1 = corner1.translate(BlockFace.TOP);
+			corner2 = corner2.translate(BlockFace.TOP);
```
megagamer410, line 120: The 3 iteration for loop, combined with these 2 lines fixes a bug in Minecraft that allows players to create 2 nether portals that are touching, causing some of both portals to become invisible. (Minecraft Bug Fix)

src/main/java/org/spout/vanilla/material/item/tool/FlintAndSteel.java
```diff
@@ -68,6 +63,11 @@ public void onInteract(Entity entity, Block block, Action type, BlockFace clicke
 						inv.addData(inv.getCurrentSlot(), 1);
 					}
 				}
+				
+				// Handle the creation of portals
+				if (VanillaMaterials.PORTAL.createPortal(target.translate(BlockFace.BOTTOM))) {
+					return;
+				}
```
megagamer410, line 70: Portal creation has to occur after the fire is placed so that the fire can be extinguished by the portal creation sequence when the obsidian frame is proper but the portal can't be generated due to a conflict with an already existing portal.

megagamer410:
Updated the commit to fix a Minecraft bug and to include W1ndwaker's suggestion.

src/main/java/org/spout/vanilla/material/block/portal/NetherPortal.java
```diff
@@ -99,6 +104,21 @@ public boolean createPortal(Block bottomBlock) {
...
 			return false;
 		}
+		// Validate that there are no portals touching the proposed one.
+		corner1 = above;
+		corner2 = above.translate(direction);
+		for (int i = 0; i < 3; i++) {
+			// Check if there is another nether portal touching the proposed one
+			for (BlockFace side : BlockFaces.NESW) {
+				// If so, don't build the current one, and extinguish the fire.
+				if (corner1.translate(side).getMaterial().equals(VanillaMaterials.PORTAL) || corner2.translate(side).getMaterial().equals(VanillaMaterials.PORTAL)) {
```
DDoS: switch:
corner1.translate(side).getMaterial().equals(...
with:
corner1.translate(side).isMaterial(...
 and I'll merge this (same thing with corner2).

PR #333, opened 2012-10-14
Fixed Chests

kitskub:
This pull WAS going to be to fix Block Components. But I'm now going to fix chests.

Fixes: http://issues.spout.org/browse/VANILLA-199

### DIFF
--- a/src/main/java/org/spout/vanilla/component/substance/material/Chest.java
+++ b/src/main/java/org/spout/vanilla/component/substance/material/Chest.java
@@ -26,22 +26,84 @@
  */
 package org.spout.vanilla.component.substance.material;
 
+import java.util.Arrays;
+import org.apache.commons.lang3.ArrayUtils;
 import org.spout.api.entity.Player;
-
+import org.spout.api.geo.cuboid.Block;
+import org.spout.api.geo.discrete.Point;
+import org.spout.api.inventory.ItemStack;
+import org.spout.vanilla.component.inventory.window.block.DoubleChestWindow;
+import org.spout.vanilla.component.inventory.window.block.SingleChestWindow;
 import org.spout.vanilla.inventory.Container;
 import org.spout.vanilla.inventory.block.ChestInventory;
+import org.spout.vanilla.inventory.block.DoubleChestInventory;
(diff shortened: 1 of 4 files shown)


windy1:
Is this actually working? Wasn't sure if block components had been implemented completely yet.

kitskub:
Well, it at least doesn't do this: http://issues.spout.org/browse/VANILLA-199

windy1:
Okay, well, I'm not sure if the implementation in SpoutAPI is even correct. Why don't we have Block implement ComponentHolder @Zidane?

kitskub:
I think it's because each block should only have one component, but ComponentHolder allows multiple.

windy1:
Why should Blocks be limited to one Component? I can't think of a reason.

Blocks should however be limited to one **Entity** I think.

kitskub:
Not sure, but that's what the current code is saying.

kitskub:
If it helps anything, I'm working on adding back in entity-->block interaction. I can add to this pull or put it in a separate one.

kitskub:
Oh, it doesn't work after a server restart. This is probably due to the entity not saving, rather than the component not being set.

Afforess:
> Why should Blocks be limited to one Component? I can't think of a reason.

Why should they have more than one? They are limited to one material.

This pull is needed anymore due to changes in how I implemented BlockComponents.

kitskub:
Ok, the vanilla pull is not needed, but I think at least part of the SpoutAPI and/or Spout pull still is. Also, shouldn't ComponentMaterial be removed?

windy1:
@kitskub @Afforess is redoing the design entirely so the Spout PRs will likely be invalid as well :/

kitskub:
@W1ndwaker me and Afforess talked. I'm going to update all the pulls to fit current specifications. Please reopen.

Afforess:
Vanilla Team: Do not pull w/o talking to me first. I am still changing Block Components even more.

kitskub:
No longer needed because of https://github.com/VanillaDev/Vanilla/commit/2aa712b29d88c1f6444ab0d734731f44ce0de6d

PR #336, opened 2012-10-30
Added the first models for vanilla blocks

Waterpicker:
Signed-off-by: Jared Johnson Waterpicker@rocketmail.com


RoyAwesome:
Please update the cube models to the .uvs file format defined here: http://forums.spout.org/threads/client-sprint-2-graphical-boogaloo.5612/page-2#post-73368

Raphfrk:
Can we have a system where models are by reference (which I think you already have).

It seems like you already have something like that?

Btw, how does it know not to draw faces then?  Ideally, if the neighbour is opaque, then it doesn't need to draw that face at all.  Maybe there needs to be a way to mark certain faces as edge faces.

Raphfrk:
I did water and stone.  You need to add files to the /resources directory too.

PR #361, opened 2012-11-07
Centering the crosshair. Re-allow client to use /damage.

Karang:


src/main/java/org/spout/vanilla/component/player/HUDComponent.java
```diff
@@ -397,7 +397,7 @@ private void initHUD() {
 		final TexturedRectComponent crosshairRect = crosshair.add(TexturedRectComponent.class);
 		crosshairRect.setRenderMaterial(iconsMaterial);
 		crosshairRect.setColor(Color.WHITE);
-		crosshairRect.setSprite(new Rectangle(-0.005f, -0.005f, 0.05f * SCALE, 0.05f));
+		crosshairRect.setSprite(new Rectangle(-0.025f * SCALE, -0.025f, 0.05f * SCALE, 0.05f));
```
Raphfrk, line 400: I made the same change, must have forgot to push,

PR #365, opened 2012-11-11
Input management in plugin

ghost:


zml2008:
Why should basic input be in Vanilla?

ghost:
Because i think these input are specific to vanilla, and the others plugins can use other input. And i need all input with a stored state in a input state to reset it when the focus is lost to prevent sticky button.

ghost:
Moved in https://github.com/VanillaDev/Vanilla/pull/368 (i changed my repo branch)

PR #372, opened 2012-11-14
Start implementing the authentification.

Karang:

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/ClientLoginAuth.java
+++ b/src/main/java/org/spout/vanilla/protocol/ClientLoginAuth.java
@@ -1,0 +1,103 @@
+/*
+ * This file is part of Vanilla.
+ *
+ * Copyright (c) 2011-2012, VanillaDev <http://www.spout.org/>
+ * Vanilla is licensed under the SpoutDev License Version 1.
+ *
+ * Vanilla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In addition, 180 days after any changes are published, you can use the
+ * software, incorporating those changes, under the terms of the MIT license,
+ * as described in the SpoutDev License Version 1.
+ *
+ * Vanilla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License,
+ * the MIT license and the SpoutDev License Version 1 along with this program.
+ * If not, see <http://www.gnu.org/licenses/> for the GNU Lesser General Public
+ * License and see <http://www.spout.org/SpoutDevLicenseV1.txt> for the full license,
+ * including the MIT license.
+ */
+package org.spout.vanilla.protocol;
+
+import java.io.BufferedReader;
+import java.io.DataOutputStream;
+import java.io.InputStreamReader;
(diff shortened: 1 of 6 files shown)


Zidane:
I want some input from the other Spout Developers before I pull this

src/main/java/org/spout/vanilla/protocol/ClientLoginAuth.java
```diff
@@ -0,0 +1,103 @@
...
+	
+	@Override
+	public void run() {
+		HttpURLConnection conn;
+		String result = "";
+		try {
+			URL url = new URL(targetURL);
+			conn = (HttpURLConnection) url.openConnection();
+			conn.setRequestMethod("POST");
+            conn.setRequestProperty("Content-Type", "application/x-www-form-urlencoded");
```
ZNickq, line 69: Formatting?

src/main/java/org/spout/vanilla/protocol/VanillaProtocol.java
```diff
@@ -129,9 +133,7 @@ public Message getKickMessage(ChatArguments message) {
 
 	@Override
 	public Message getIntroductionMessage(String playerName) {
-		//return new PlayerHandshakeMessage(VanillaPlugin.MINECRAFT_PROTOCOL_ID, playerName); //TODO Fix this Raphfrk
-
-		return null;
+		return new PlayerHandshakeMessage(VanillaPlugin.MINECRAFT_PROTOCOL_ID, playerName, "127.0.0.1", 25565); //TODO Fix this Raphfrk
```
ZNickq, line 136: Is this TODO still needed? looks like you fixed it :P

src/main/java/org/spout/vanilla/protocol/ClientLoginAuth.java
```diff
@@ -0,0 +1,103 @@
...
+
+		} catch (Exception e) {
+			e.printStackTrace();
+		}
+		
+		System.out.println("Result: "+result);
+		
+		if (result.contains(":")) {
+			String sessionId = result.split(":")[3].trim();
+			session.getDataMap().put(VanillaProtocol.SESSION_ID, sessionId);
```
Raphfrk, line 98: I am not sure authing for the client needs to have its own thread?  It's not like the client will be doing anything at the time. 

Anyway, if you are doing it this way, setting the data map isn't enough.  You need to ping the main login thread for it to wake up.

windy1, line 98: It's a Runnable @Raphfrk ;)

Raphfrk, line 98: Hmm, ok, but what not just method calls then?

windy1, line 98: I don't think it's been fully implemented.

Zidane, line 98: I'm not seeing the need for a Runnable either.

We can simply suspend the client until they pass the login screen no?

Karang, line 98: The things is that it needs to be called when the client create the player's session, before sending the handshake message, just after the player typed the connect command. So it's to avoid client's freezing while doing the request. Or if we store the id outside of the session, we can do as you said and get it at the launch.

src/main/java/org/spout/vanilla/protocol/VanillaProtocol.java
Karang, line 136: right :)

src/main/java/org/spout/vanilla/protocol/ClientLoginAuth.java
Zidane, line 98: Session has access to the datatable so storing is possible. If you can make the change to that (I think it would be better) then i can pull.

Either way it can't be auto merged for some reason.

PR #373, opened 2012-11-15
Add Light color via a BatchEffect

ghost:
Add /torch command to place a torch
BatchEffect for light added in all vanilla material


src/main/java/org/spout/vanilla/command/TestCommands.java
```diff
@@ -117,6 +117,31 @@ public void traceray(CommandContext args, CommandSource source) throws CommandEx
...
+			player = (Player) source;
+		} else {
+			player = ((Client)Spout.getEngine()).getActivePlayer();
+		}
+		
+		BlockIterator blockIt = player.get(HitBlockComponent.class).getAlignedBlocks();
+		
+		Block block = null;
+		while (blockIt.hasNext()) {
+			block = blockIt.next();
```
Raphfrk, line 137: I think there might be a method in the BlockIterator that does this (finds the first hit block)?

PR #375, opened 2012-11-16
Add OpenGL dependance

ghost:
Add Light with BatchEffect && RenderEffect (not fully working)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,7 @@
 		<protocolVersion>49</protocolVersion>
 		<buildNumber>0</buildNumber>
 		<powermock.version>1.4.12</powermock.version>
+		<lwjglVersion>2.8.3</lwjglVersion>
 	</properties>
 
 	<!-- Organization -->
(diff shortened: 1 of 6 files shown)


zml2008:
@RoyAwesome I believe we didn't want LWJGL in Vanilla/outside of engine. Right?

ghost:
If we you want support for gl 11, i don't see other way.
Just careful to not call class with opengl on server.

windy1:
I was under the impression that we wanted everything you could need for rendering to be in Spout and not ever need plugins to depend LWJGL. ( @Olloth )

Salynne:
After a discussion on IRC last night or the night before with roy it was determined that this is impossible if we want to allow render effects with gl11 support to exist.

That said, you still don't want lwjgl on a server and the discussion we were having windwaker involved depending on lwjgl solely for the keyboard enum in a server plugin.

windy1:
Impossible even if we have a wrapper for it in SpoutAPI?

Salynne:
Wrap all of G11? I mean if you want to...I guess you can. None of the sane people are going to attempt it.

windy1:
@Olloth well no, I thought this was for one feature but I guess I am mistaken.

ghost:
How do you want allow plugin to make custom render effect, with GL 11, if you don't allow to depend of LWJGL ?

Ok for gl 2 & 3, we can do everything by giving data to shaders.

So, you must decide, depend of LWJGL in plugin or drop GL 11 support

Impossible even if we have a wrapper for it in SpoutAPI? -> NO NO NO NO NO

ghost:
I have a few classes wrapping around the basic GL(version) classes and the GLU class if you want it.

PR #376, opened 2012-11-17
Add a listing of the living components

Snaipe:
Basically, this will allow living components to be looked up from their names, and to cycle through them.


Snaipe:
Never mind that, the exception message is incomplete ( missing a simple quote and space )

PR #377, opened 2012-11-17
Add a listing of the living components

Snaipe:
Basically, this will allow living components to be looked up from their names, and to cycle through them.

### DIFF
--- a/src/main/java/org/spout/vanilla/component/living/VanillaLivingComponents.java
+++ b/src/main/java/org/spout/vanilla/component/living/VanillaLivingComponents.java
@@ -1,0 +1,114 @@
+/*
+* This file is part of Vanilla.
+*
+* Copyright (c) 2011-2012, VanillaDev <http://www.spout.org/>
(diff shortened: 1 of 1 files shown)


windy1:
This isn't really that useful, I would pull this if the classes were in the CreatureType and ObjectType enums. :/

Snaipe:
@W1ndwaker 
Well, I do find it useful, mostly for command processing. Having a lookup from the name would be more convenient than making a whole bloc of conditions or putting the same component listing inside the command class.
I could give two examples here : the /spawnmob <entity-type> command, and the /remove <entity-type> command. Both take the component name in parameter, and could be used in a lookup.
Also, where are the CreatureType and ObjectType enums you mentioned ? I can't seem to find them.

Edit : oh, found them. I see what you meant then. I should probably put the classes in those enums rather than in an other separated file. Fair enough, I'll be doing an other PR for that.

windy1:
Very nice :)

PR #378, opened 2012-11-17
Add the component for each creature in CreatureType

Snaipe:
This will bind to each creature descibed in the CreatureType enum its component.

I will do the same thing in an other PR for the ObjectType enum, as for now, there are no components for ENDER_PEARL, WITHER_SKULL, ITEM_FRAME, EXP_BOTTLE.

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/entity/creature/CreatureType.java
+++ b/src/main/java/org/spout/vanilla/protocol/entity/creature/CreatureType.java
@@ -26,42 +26,79 @@
  */
 package org.spout.vanilla.protocol.entity.creature;
 
-public enum CreatureType {
-	BAT(65),
-	BLAZE(61),
-	CAVE_SPIDER(59),
-	CHICKEN(93),
-	COW(92),
-	CREEPER(50),
-	ENDER_DRAGON(63),
-	ENDERMAN(58),
-	GHAST(56),
-	GIANT(53),
-	IRON_GOLEM(99),
-	MAGMA_CUBE(62),
-	MUSHROOM_COW(96),
-	OCELOT(98),
-	PIG(90),
-	PIG_ZOMBIE(57),
-	SHEEP(91),
-	SILVERFISH(60),
-	SKELETON(51),
-	SLIME(55),
-	SNOW_GOLEM(97),
-	SPIDER(52),
-	SQUID(94),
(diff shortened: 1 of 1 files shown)


windy1:
No ObjectType? :P

Karang:
We should do an entity prefab for each one of them.

Snaipe:
@W1ndwaker I would gladly have done that if that was possible :<
The components for WITHER_SKULL, ITEM_FRAME, ENDER_PEARL, and EXP_BOTTLE are missing atm...

src/main/java/org/spout/vanilla/protocol/entity/creature/CreatureType.java
```diff
@@ -26,43 +26,56 @@
  */
 package org.spout.vanilla.protocol.entity.creature;
 
+import org.spout.vanilla.component.living.LivingComponent;
+import org.spout.vanilla.component.living.hostile.*;
```
windy1: No star imports please.

src/main/java/org/spout/vanilla/protocol/entity/creature/CreatureType.java
```diff
@@ -26,43 +26,56 @@
...
+	private CreatureType(int id, Class<? extends LivingComponent> component) {
 		this.id = id;
+		this.component = component;
 	}
 	
 	public int getId() {
 		return id;
 	}
+	
+	public Class<? extends LivingComponent> getComponent()
```
windy1: Formatting:

``` java
public Class<? extends LivingComponent> getComponent {
    return component;
}
```

Snaipe:
@W1ndwaker 
Here goes, all fixed.

Zidane:
I agree, @W1ndwaker these should be prefabs. I want to move towards that system.

Snaipe:
@Zidane 
you mean entities with those components ? I could do a getEntity() method which creates a new Entity when called, attach the component, then returns the entity.
Edit: I just realized that there's no way to make such a method without parameters. At least it should take a Point or a Transform..

windy1:
This needs to be rebased.

Snaipe:
Indeed. I have some work to do with entity prefabs before though.

Snaipe:
Rebased it. I will link the entries to the prefab in an other commit.

PR #379, opened 2012-11-18
New OBJ files for stairs and wooden fences

SelennLamson:
Fences :
Connections are indicated by N-E-S-W

Stairs :
I -> reversed stairs
Normal stairs go down in the indicated direction (N-E-S-W) and
reversed stairs go up in these directions.


DDoS:
Is it really necessary to have separate OBJ files for stone, brick, cobble, wood and sandstone stairs? Don't they all have the same models? The textures and properties are different, but not the geometry...

SelennLamson:
I asked L5D and Karang who said me to make all textures and orientations. So I did.

I made a soft which allows me to change quickly a texture applied on an obj file, so it's easy for me.
If you can use only one model for all its colors and orientation, you just have to delete the other ones ^^ but they're here in case you can't.

SelennLamson:
Ow, I thank :
It may be very very hard to change orientations in the code. Changing texture is easy (I made it in one hour, without having programing for two months). But the fact when rotating the block is that some texture have not to rotate.
Explanations ? The top and bottom texture have to be in adequacy with neighbor blocks of same kind (cobblestone stairs and cobblestone blocks, for example). For the bottom of the block, it could be done without difficulties. But for the top, with the two steps of the stairs, I don't know if it's as easy as you think.

DDoS:
I'm still not sure I understand why you can't use only one set of models for all types of stairs...

SelennLamson:
The bottom of a texture is always directed to the South (on top and bottom faces). Even on a stair.
Imagine a North orientated stair (with cobblestone). On its top face, bottom of the texture is fully on the top step.
Now consider a West orientated stair (still with cobblestone). This time, the bottom of the texture on its top face is half one the bottom step, and half on the top step.
Could you do that change easily in the code ?

PR #384, opened 2012-11-20
entities obj : human - zombie - creeper

SelennLamson:
Human was done but strangly, i did it again.
I did the zombie/pigzombie the same as human, but with horizontal arms.
Don't know if it's usefull or if you prefer to use human model for all
of them.
Creeper is done.


windy1:
@Tomlabete Is it possible we could get some screenshots from these in Blender? I would love to see how they look :)

PR #397, opened 2012-11-25
Add snow levels in code + Edit camera speed

SelennLamson:
Snow levels are ready to be use. But in game, there's always the whole
snow block instead of 7 levels. To fix.

Camera speed multiplied by 10 for the tests.

PS : didn't know how to make 2 PR with one commit, so I did one with the two commits.

Signed-off-by: Thomas LAMSON tomlabete@numericable.fr


windy1:
For the future, everything you need to know about squashing commits is here: http://spout.in/pr

PR #406, opened 2012-12-08
tall grass obj

Meiun:
This is a correction for the one sent over IRC. 
Had asked if you guys needed it to have texture maps and someone said no so didn't include VT indices or VT face definitions.
This has both added. In MC this has some offset variations and is not centred like this. Can create off centre versions if needed.


DDoS:
@Zidane added the same file in one of his commits. I'm going to close this.

Also this doesn't match PR guidelines.

PR #415, opened 2012-12-13
Implemented BlockChangeCause for Form, Grow, Ignite and Remove

dredhorse:
Form =  Spread_Random, Spread_From, Forming
Grow
Ignite = Lava, Flint_and_Steel, Spread,  Lightning, Fireball
Remove = Fade, Melt, Disappear, Decay

all of those causes for the BlockChangeEvent still need to be implemented.

### DIFF
--- a/src/main/java/org/spout/vanilla/event/cause/BlockFormCause.java
+++ b/src/main/java/org/spout/vanilla/event/cause/BlockFormCause.java
@@ -1,0 +1,96 @@
+/*
+ * This file is part of Vanilla.
+ *
+ * Copyright (c) 2011-2012, VanillaDev <http://www.spout.org/>
+ * Vanilla is licensed under the SpoutDev License Version 1.
+ *
+ * Vanilla is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In addition, 180 days after any changes are published, you can use the
+ * software, incorporating those changes, under the terms of the MIT license,
+ * as described in the SpoutDev License Version 1.
+ *
+ * Vanilla is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License,
+ * the MIT license and the SpoutDev License Version 1 along with this program.
+ * If not, see <http://www.gnu.org/licenses/> for the GNU Lesser General Public
+ * License and see <http://www.spout.org/SpoutDevLicenseV1.txt> for the full license,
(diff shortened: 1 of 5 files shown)


src/main/java/org/spout/vanilla/event/cause/BlockFormCause.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Caused when a block is formed or spreads
+ * todo needs to be implemented
+ */
+public class BlockFormCause extends BlockChangeCause {
+	/**
+	 * The different causes why a Block is formed in the world.
+	 */
+	public enum FORM_CAUSE {
```
windy1: This should be on a new line.

src/main/java/org/spout/vanilla/event/cause/BlockIgniteCause.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Caused when a block is ignited
+ * todo needs be implemented
+ */
+public class BlockIgniteCause extends BlockChangeCause {
+	/**
+	 * The different causes why a Block was ignited.
+	 */
+	public enum IGNITE_CAUSE {
+		/**
```
windy1: And here

src/main/java/org/spout/vanilla/event/cause/BlockFormCause.java
dredhorse: the { ?

windy1: No the first enum value. "FORM_CAUSE"

dredhorse: that IS the enum...

kitskub: W1ndwaker, that's not an enum value. Dredhorse: FormCause

src/main/java/org/spout/vanilla/event/cause/BlockIgniteCause.java
kitskub: Same as above

src/main/java/org/spout/vanilla/event/cause/BlockRemoveCause.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Caused when a block fades, melts, disappears or decays
+ * todo needs to be implemented
+ */
+public class BlockRemoveCause extends BlockChangeCause {
+	/**
+	 * The different causes why a Block is removed from the world.
+	 */
+	public enum REMOVE_CAUSE {
```
kitskub: Same

src/main/java/org/spout/vanilla/event/cause/BlockFormCause.java
windy1: Oh herp, I just woke up, but anyways, enum names should never be all caps, name them as you would classes.

dredhorse: ok.. will do in the other too... good morning :-)

PR #417, opened 2012-12-14
Implementing / Refactoring Events

dredhorse:

Before pulling please discuss: http://forums.spout.org/threads/cause-and-event-system.6129


src/main/java/org/spout/vanilla/event/block/FurnaceSmeltEvent.java
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Sets the result ItemStack
+	 * @param newResult the result ItemStack
+	 */
+	public void setResult(ItemStack newResult) {
+		result = newResult;
+	}
+
+	/**
+	 * Returns the {@link org.spout.vanilla.event.cause.misc.FurnaceSmeltCause} which caused the FurnaceSmeltEvent
```
kitskub: Nah-ah! Look at the construtor. Also look t the javdocs for the class, same as FurnaceBurnEvent

dredhorse: yeah  copy paste error... will fix that if this is the way we go further, thanks for catching it...

dredhorse:
I just found out that my furnace causes suck... because a furnace doesn't has any reference to the block he is, so I will change that in a sec too.

kitskub:
@dredhorse : getOwner()?

dredhorse:
sigh.. you are right.. thanks

windy1:
It doesn't have a reference to the block, just the position, and that is intentional.

PR #420, opened 2012-12-15
Spawn Eggs spawn again, Sheeps eat grass.

timmhirsens:
Spawn eggs spawn entities again.
Some don't work, but this should be fine for testing.
Sheep also eat grass again, ported the change I made a few months ago to the component system.

Signed-off-by: fr1zle timmhirsens@gmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/material/item/misc/SpawnEgg.java
+++ b/src/main/java/org/spout/vanilla/material/item/misc/SpawnEgg.java
@@ -26,44 +26,69 @@
  */
 package org.spout.vanilla.material.item.misc;
 
+import org.spout.api.component.Component;
 import org.spout.api.entity.Entity;
 import org.spout.api.event.player.PlayerInteractEvent.Action;
+import org.spout.api.geo.LoadOption;
 import org.spout.api.geo.cuboid.Block;
 import org.spout.api.material.Material;
 import org.spout.api.material.block.BlockFace;
-
+import org.spout.vanilla.component.living.hostile.Blaze;
+import org.spout.vanilla.component.living.hostile.CaveSpider;
+import org.spout.vanilla.component.living.hostile.Creeper;
+import org.spout.vanilla.component.living.hostile.Ghast;
+import org.spout.vanilla.component.living.hostile.MagmaCube;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spout/vanilla/component/misc/SheepEatingGrassComponent.java
```diff
@@ -0,0 +1,100 @@
```
windy1: This can just go in the Sheep's component since this will only be attached to sheep ;)

timmhirsens: I don't really know how the component system works, but wouldn't this then make the sheep tick only if it can eat grass?

windy1: Sheep has an onTick too, so all of the tick logic can go in Sheep.

timmhirsens: so the canTick logic goes into onTick?

windy1: Yeah it can, basically canTick is just structured like this

``` java
if (canTick()) {
   tick();
}
```

every tick

timmhirsens:
Refactored, as Windwaker suggested.

src/main/java/org/spout/vanilla/material/item/misc/SpawnEgg.java
```diff
@@ -71,6 +96,6 @@ public void onInteract(Entity entity, Block block, Action type, BlockFace clicke
 		if (type != Action.RIGHT_CLICK) {
 			return;
 		}
-		//TODO re-write spawn egg spawning handling.
+		block.getWorld().createAndSpawnEntity(block.translate(clickedface).getPosition(), entityComponent, LoadOption.LOAD_GEN);
```
Zidane: This shouldn't generate a chunk nor load one as you would be in the chunk itself and it would be loaded.

timmhirsens: Fixed.

PR #423, opened 2012-12-18
overhauled torches for greater MC parity

Meiun:
Signed-Off-By: Hung-Lin <nymphalis.aloft@gmail.com

the previous ones might not work with all texture packs as well.
not sure, dont want to risk it.


DDoS:
There's an issue with vertex winding order. Some exterior faces are being culled.

PR #429, opened 2012-12-19
gl didnt like my quads; triangulated and fixed.

Meiun:
clockwise quads were being rendered as if they were made of one
clockwise and one counter clockwise tri.
didn't notice this on main machine, but was apparent on another machine.

Signed-off-by: Hung-Lin nymphalis.aloft@gmail.com


Zidane:
I can't pull this :/...its not updated to head I believe

DDoS:
I got the fixed OBJs from Meiun and I'll be adding them in manually by request.

PR #431, opened 2012-12-20
Add HealthChangeReason.PLUGIN

kitskub:
SpoutAPI: https://github.com/SpoutDev/SpoutAPI/pull/469
Spout: https://github.com/SpoutDev/Spout/pull/2146


Zidane:
Can't be pulled

kitskub:
Hmmm

kitskub:
Updated :)

windy1:
I don't see the need for this, I think plugin causes can just go under UNKNOWN

kitskub:
But, it's not unknown. It's a plugin.

dredhorse:
I think this is useful, we also have a plugin cause in spout api.

PR #440, opened 2012-12-22
implemented BlockExperienceEvent, BlockBreakEvent, minor fix to BlockFromEvent

dredhorse:


Afforess:
I'm not sure we need a block break event. We already have a block change event, and you can just check the cause to see if it was from a player breaking the block.

dredhorse:
but a block break drops experience, a block change event doesn't

Afforess:
Hmm, good point. I need to think about it then.

windy1:
@Afforess @dredhorse Can't we include the exp data in the Cause of the BlockChangeEvent?

dredhorse:
we agreed to not do that, I had the suggestion to drop a lot of the events and pass causes with all necessary data into it.

The problem is two fold:

a) Figuring out what the event was for is a pita as you need to do instance off and casting
b) you need to get the cause, set the new value and afterwards set the cause again. (or have a mutable cause object). But the cause in Events is final.

while I do agree that the event system is really powerful and can take a lot of plugins listening to a very generic event, I agree with @Sleaker and @Xemsdoom that afterwards figuring out what the event was for isn't a nice solution.

So I tried to implement a mixture between the basic spout events and what is available in bukkit atm, combining events which make sense and otherwise having separate events.

The only way I can think of a system with more generic events (I made a PR for that in the beginning for the furnace events) is to use the cause system fully to transfer all needed data, have an enum perhaps which defines the cause on top and than have a massivly over coded event which has getters for all the different causes.

so you could do something like this:

getCauseType

switch (causeType)
      FurnaceSmeltEvent: 
                    event.getFurnaceSmeltCause....
                           etc.

that wouldn't require tons of instance off and casting on the developers side, and it would also allow easy extension from the vanilla side.

but it would hamper extension from the plugins side as the CauseTypes would be defined in the enum (well we could always do an UNKNOWN there too and keep the getCause like it is atm too.

I honestly think that the whole Vanilla Team should work on bukkitbridge as it shows what stuff we are missing, not only the stuff we know but also the tiny little bits, like the Experience drop when you break blocks.

Afforess:
@W1ndwaker - causes are immutable snapshots for why an event happened, not storage containers. 

@Dredhorse - What if we have just a ExperienceChangeEvent that is fired when players gain or lose xp. To show that xp gain was from breaking a block, you can reuse the PlayerDiggingCause that already exists. The actual amount of xp gain or lost would still be in the event.

dredhorse:
hmm... that would be behind the "BlockBreakEvent"... so if you want to change the xp being "dropped" from block break you would use that, but if  you would like to cancel the break you would use the first.

that means that you would have to listen to two events, BlockChangeEvent being quite complicated to track because of casting and instanceOf, just to be able to disallow breakage AND manage xp drop depending on "skills" for example.

I see this from the plugin developers point of view... less events are more complicated to use atm, sure you only need to hook one event and can do all kind of stuff with it, but you need a quite big if... instanceOf... cast structure if you want to use it.

more detailed event can break down the classes and make it simpler.

On the other side is the amount of work to go into creating the events, but honestly only atm this is a lot of work, it also probably helps later on when mojang changes stuff in the background, like xp dropping for breaking blocks.

don't know... what do other think about it from the team? I know that @Zidane was also for using as few events as possible... some people from the community @Sleaker for example wasn't really pro it.

BTW: We should probably have that ExperienceChangeEvent never the less :-)

On a site note: Merry Chirstmas

windy1:
@dredhorse I think a good goal to shoot for is as least events as possible without it becoming ambiguous.

Afforess:
@dredhorse - the goal is to keep separate categories of events separate, and keep things as concise as possible.

XP is different from breaking blocks. You can get XP from killing monsters, or from plugins, etc. So why name an event **BlockBreakEvent** if the point is to allow modification of XP? Why not have an _ExperienceChangeEvent_ instead?

Some of the confusion may be coming from the fact that naming relies heavily on English, so there is a language barrier. Also, what I am trying to achieve is more an art than a science, so there is no wrong answer, just good and better answers.

dredhorse:
Like I said, we should probably go with the ExperienceChangeEvent, but also with the BlockBreakEvent.

I agree that the BlockXPEvent is probably not a good idea. I still think though that going the BlockChangeEvent in the way it is atm isn't such a good idea, as the code to handle the different real events afterwards is awkward.

Even if you implement a PlayerBlockBreakCause, you will need to get the cause, instanceOf it to figure out if it is really a PlayerBlockBreakCause, cast it and than go further.

If you don't do that, figureing out what really happens in the BlockChangeEvent becomes even harder and requires a lot more documentation or HowTos.

But... we can always change it later on.

luca-moser:
I am really against generic events. As mentioned above, a developer would have to check if the cause of the event, is what he wants it to be and this actually makes a performance loss:
If we have a BlockChangeEvent and a plugin only needs to execute some code when a block was placed, it is unnecessary that the EventHandler is going to call each time that listener which only wants to know about the block placement.

```
    public void onBlockPlace(BlockChangeEvent event){

        Cause<?> someCause = event.getCause();

        // Return out of this, I don't need anything done if a player doesn't place a block
        if(someCause != BlockChangeEvent.PLAYER_BLOCK_PLACE){
            return;
        }

        // Code I want to be executed on player block placement
    }
```

(I know that the upper doesn't really represent any real code of how it will be)

I don't think there is a benefit of generic events, as I don't really see having less events being more performant and it isn't more developers friendly in my eyes.

Afforess:
If your main complaint is performance, then get over it. Events are extremely marginal in terms of performance.

I'm in favor of more generalized events because Bukkit did specific events so badly. Want to listen to all block changes on a server? Enjoy listening to BlockBreakEvent, and EndermanPlaceEvent, and EntityExplosionEvent, and BlockFlowEvent, and... and you can't even cover all the cases, plugins that change blocks never even trigger an event. So screw that model. Generic Events that catch all cases are much, much, better.

luca-moser:
How are they much much better, if a developer needs to check if the specific cause or "subevent" of that event is, what he wants to listen to. I don't see how less events somehow benefit to the developer, since it is annoying to create if statements, just because the event is meant generic.
Performance? Well, you should always have that in mind and in this case it even makes things easier aftwards. I know its annoying to go make X number of events in Spout or Vanilla but afterwards you can enjoy the fact, that you are exactly listening to what you want to listen to. 
If a plugin doesn't call an event when it does something, well, then it is the plugin's fault.
I am fairly aware that event system are somtimes the slowest part of software but we don't need to make it even less performant.

Afforess:
> I am fairly aware that event system are somtimes the slowest part of software but we don't need to make it even less performant.

Did you misread what I said? Events do not contribute to any of the performance issues in the server. File I/O is the number one performance issue. Events do not even show up in any of the profilers I've used.

Feel free to keep beating a dead horse, but don't expect me to change my opinion.

Zidane:
@dredhorse 

Do you need to do anymore work to this? Am I okay to review in prep for a pull?

dredhorse:
sorry... I was sick... slept around 36 hours in the last 2 days.

@Afforess, @XemsDoom, @Sleaker and @Zidane 

what about a middle ground? 

using BlockChangeEvent as the master and use the other events as childs... which means the other events extend the BlockChangeEvent?

I''m not looking at the performance aspect, as we know that this isn't important. 

I'm looking at ease of documentation and usage of the events. Only having the BlockChangeEvent for everything is a documentation nightmare for use cases and implementation wise also not very comfortable. But it would be helpful for certain plugins to be able to track all changes to a block at one place.

Concerning the pull itself, I will redo it to use the suggested Experience Change event (a lot better I think like this approach).

kitskub:
@dredhorse that middle ground is pretty much perfect. But, to make sure: if an event specified in an EventHandler is super to the class being called, it'll be called, correct? If so, it's the best of both: plugins can listen for specific or broad.

Zidane:
@dredhorse 
You have a go. Make it so.

Sleaker:
that's what I was recommending earlier. Allow plugins to hook generic events or specific ones. have seperate handlers for both, but make sure upstream listeners receive the downstream events.

luca-moser:
I am totally fine with subevents together with generic events, as this would resolve the problem of annoying Cause check and is much cleaner. Less events != better system

Afforess:
> using BlockChangeEvent as the master and use the other events as childs... which means the other events extend the BlockChangeEvent?

Well you aren't really going to fix things then, because BlockChangeEvent is currently fired in the Engine. If you fire it other places, you will cause 2 block change events for 1 block change, which is bad. I'm not seeing a way you are going to achieve your middle ground, which is why causes exist in the first place.

kitskub:
Looks like we have an intersection with the EventHandler stuff (http://forums.spout.org/threads/order-monitor.6223/). Perhaps there's a possibility to kill two birds with one stone here.

Zidane:
@Afforess @Sleaker @dredhorse 

@zml2008 Added support for child handlers. This solves the issue no?

dredhorse:
@Afforess I came to that conclusion too after 1 hour of tf2 and did lie awake for 2 hours thinking about a solution.

I think I have an idea besides the Child Handlers (will look into that too). Let me quickly code it up as another PR to take a look I think that COULD be a solution and atm I think it might be a big enhancement.

dredhorse:
ok... idea is up... https://github.com/VanillaDev/Vanilla/pull/460

now looking into the commit zml did

dredhorse:
looks like the commit from @zml2008 would solve the issue Afforess came up with, I would STILL like to get feedback for my other idea. see also http://forums.spout.org/threads/event-cause-system.6343/

ghost:
cause event system?
how about:
@EventHandler(cause = BlockEvent.Cause.EXPLOSION)
public void onEvent(BlockBreakEvent ev)

to limit the ifs if someone just want to handle one cause?

zml2008:
That's not possible @macer1. And what does everybody have against if
statements anyway?

windy1:
Yeah I'm with @zml2008 on this, a few `instanceof` never hurt anyone, and it's a helluva lot better then some convoluted system. K.I.S.S

dredhorse:
so any comment about the other pr and the link I posted?

luca-moser:
I totally like @dredhorse's event handling

PR #447, opened 2012-12-24
Fixed health being stuck at 1. Also made the /hunger work in server mode.

greatman:
Signed-off-by: William Lebel lebelwilliam@gmail.com


src/main/java/org/spout/vanilla/component/living/neutral/Human.java
```diff
@@ -70,7 +70,7 @@ public void onAttached() {
 		//		if (holder instanceof Player) {
 		//			((Player) holder).teleport(holder.getTransform().getPosition().add(0, 1.85F, 0));
 		//		}
-		if (getAttachedCount() < 1) {
+		if (getAttachedCount() == 1) {
```
Zidane, line 73: This would mean the attached count was never set properly.

Fix that instead?

PR #451, opened 2012-12-27
hideEntity and showEntity

kitskub:

SpoutAPI: https://github.com/SpoutDev/SpoutAPI/pull/481
Spout: https://github.com/SpoutDev/Spout/pull/2151


src/main/java/org/spout/vanilla/VanillaListener.java
```diff
@@ -137,4 +139,14 @@ public void onBlockChange(BlockChangeEvent event) {
 			}
 		}
 	}
+	
+	@EventHandler
+	public void onEntityHide(EntityHideEvent event) {
+		event.getHiddenFrom().get(PlayerListComponent.class).force();
```
Zidane: Should check if its a player first. I don't want to do an un-necessary ping.

PR #452, opened 2012-12-27
Add PlayerToggleSneakingEven and PlayerToggleSprintingEvent

kitskub:

### DIFF
--- a/src/main/java/org/spout/vanilla/event/player/PlayerToggleSprintingEvent.java
+++ b/src/main/java/org/spout/vanilla/event/player/PlayerToggleSprintingEvent.java
@@ -1,0 +1,68 @@
+/*
+ * This file is part of Vanilla.
+ *
+ * Copyright (c) 2011-2012, Spout LLC <http://www.spout.org/>
+ * Vanilla is licensed under the Spout License Version 1.
+ *
+ * Vanilla is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option)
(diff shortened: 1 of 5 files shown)


src/main/java/org/spout/vanilla/event/player/PlayerToggleSprintingEvent.java
```diff
@@ -0,0 +1,67 @@
...
+	@Override
+	public void setCancelled(boolean cancelled) {
+		super.setCancelled(cancelled);
+	}
+
+	/**
+	 * Get the GameMode the player will have.
+	 * @return The GameMode the player will have
+	 */
+	public boolean isSprinting() {
```
windy1: You should be able to set this. (Same for sneaking)

Zidane: ^ plus your javadocs are wrong.

kitskub: Why should you be able to set it? Just cancel it?

Sleaker:
Are you sure this compiles? PlayerEvent isn't cancellable and doesn't have the method setCancelled() I'm not sure how this works.  Events that implement cancellable need to actually have isCancelled defined within them.  You can't call a super method that doesn't exist and it's not explicitly defined in PlayerEvent, or in Cancellable (interface).

kitskub:
Yes it compiles. Check EVERY other cancellable event if you don't believe me.

Sleaker:
Ah I see, the code is boxed into Event... odd.

PR #454, opened 2012-12-29
Fix two issues encountered when cancelling a PlayerBreakCause event (via BlockChangeEvent)

Hidendra:
This is cross-referenced with the PR on [Spout](https://github.com/SpoutDev/Spout/pull/2154)

There are currently two problems within Spout and Vanilla related to cancelling block break events. One is minor and the second is subtle yet a problem. These are listed below.
- Spout: Currently there are two main stages during the life cycle of a block being being broken. `onDestroy()` and `onPostDestroy()`. In Spout, `onDestroy()` will destroy the block. In Vanilla, `onPostDestroy()` generates drops. These two stages are entirely independent of each other so even if the block break is cancelled in `onDestroy()` the method `onPostDestroy()` is still called and thus drops are still generated. This was doubly confirmed by _always_ cancelling a `BlockChangeEvent` with a `PlayerBreakCause` cause and then breaking a diamond block. The block (albeit destruction cancelled) constantly dropped diamonds, many diamonds.
- Vanilla: When a block break is cancelled (via `BlockChangeEvent`) the block is not sent back to the client, so it will appear broken until it is forcibly updated by the server or you log out and in.

I was able to fix both of these without making any API breaking changes by moving the creation/call of the `BlockChangeEvent` from `SpoutChunk#setBlockMaterial` to the digging packet handler in Vanilla. This fixes both issues:
- Spout: `onDestroy()` and thus `onPostDestroy()` are only ever reached if the `BlockChangeEvent` is not cancelled.
- Vanilla: Vanilla is required to send a `BlockChangeMessage`. This allows the block to be updated on the client with the correct one.

I have not throughly investigated all possible implications of these changes however do not foresee issues from a change so easy to test.

Likely this could be implemented in a better way by modifying the API itself but this is certainly one way :-)

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/handler/player/PlayerDiggingHandler.java
+++ b/src/main/java/org/spout/vanilla/protocol/handler/player/PlayerDiggingHandler.java
@@ -33,6 +33,7 @@ import org.spout.api.Spout;
 import org.spout.api.chat.style.ChatStyle;
 import org.spout.api.entity.Player;
 import org.spout.api.event.Result;
+import org.spout.api.event.block.BlockChangeEvent;
 import org.spout.api.event.player.PlayerInteractEvent;
 import org.spout.api.event.player.PlayerInteractEvent.Action;
 import org.spout.api.geo.Protection;
@@ -43,6 +44,7 @@ import org.spout.api.inventory.ItemStack;
 import org.spout.api.material.BlockMaterial;
 import org.spout.api.material.basic.BasicAir;
 import org.spout.api.material.block.BlockFace;
+import org.spout.api.material.block.BlockSnapshot;
 import org.spout.api.math.Vector3;
 import org.spout.api.plugin.services.ProtectionService;
 import org.spout.api.protocol.MessageHandler;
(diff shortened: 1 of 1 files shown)


dredhorse:
nice catch @Hidendra we need for sure test all the events throughly.. atm a lot of them aren't even hooked up or are still changing.

zml2008:
Your BlockChangeMessage sending should be fine. I had Glowstone doing that
but it must not have made it over.

Afforess:
This is a really good catch, but not an ideal solution. Firing the block change event in Vanilla means that the setMaterial call below it will also fire a block change event again, creating 2 events for only 1 actual change.

The ideal solution is to fix onDestroy so that it does not propagate a call to onPostDestroy if the event is cancelled.

Afforess:
Ok, I should have corrected the problem - I ran a test plugin that did the same as yours, cancelled diamond block events, and I did not see any diamond drops and the block remained intact. However, double check and report back if it is or is not fixed.

Relevant Commits:
https://github.com/SpoutDev/Spout/commit/d49e2128032ba61f3adae46a82707875861f5bbd
https://github.com/SpoutDev/SpoutAPI/commit/16b7b5ac670821bb8466386cfaa8a8d9739af062
https://github.com/VanillaDev/Vanilla/commit/945e160d9254403ead74f2d8d1fec439c9ea9bff

Hidendra:
Working well for me. I agree that modifying the API to support it is a better way to do it as you've done. Thanks @Afforess

windy1:
@Hidendra That's the benefit of not having too many invested developers yet ;)

PR #456, opened 2012-12-29
Updated output of commands.

Grinch:

### DIFF
--- a/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
+++ b/src/main/java/org/spout/vanilla/command/AdministrationCommands.java
@@ -70,17 +70,30 @@ public class AdministrationCommands {
 			if (!(source instanceof Player)) {
 				throw new CommandException("You must be a player to clear your own inventory.");
 			}
-			((Player) source).get(PlayerInventory.class).clear();
+			PlayerInventory inv = ((Player) source).get(PlayerInventory.class);
+			if (inv == null) {
+				source.sendMessage(plugin.getPrefix(), ChatStyle.RED, "You have no inventory!");
+				return;
+			}
+			inv.clear();
 		}
 		if (args.length() == 1) {
 			Player player = args.getPlayer(0, false);
-			player.get(PlayerInventory.class).clear();
-			player.sendMessage(ChatStyle.RESET, plugin.getPrefix(), ChatStyle.BRIGHT_GREEN, "Your inventory has been cleared.", ChatStyle.RESET);
+			if (player == null) {
+				source.sendMessage(plugin.getPrefix(), ChatStyle.RED, "Player is not online!");
+				return;
+			}
+			PlayerInventory inv = player.get(PlayerInventory.class);
+			if (inv == null) {
+				source.sendMessage(plugin.getPrefix(), ChatStyle.RED, "Player has no inventory!");
+				return;
+			}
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -145,10 +159,10 @@ public void deop(CommandContext args, CommandSource source) throws CommandExcept
 		String playerName = args.getString(0);
 		OpConfiguration ops = VanillaConfiguration.OPS;
 		ops.setOp(playerName, false);
-		source.sendMessage(ChatStyle.YELLOW, playerName, " had their operator status revoked!");
+		source.sendMessage(ChatStyle.RESET, plugin.getPrefix(), playerName, ChatStyle.RED, " had their operator status revoked!", ChatStyle.RESET);
```
zml2008: You don't need to wrap these in reset, they're set to default style already

src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -70,17 +70,30 @@ public void clear(CommandContext args, CommandSource source) throws CommandExcep
...
+				source.sendMessage(plugin.getPrefix(), ChatStyle.RED, "You have no inventory!");
+				return;
+			}
+			inv.clear();
 		}
 		if (args.length() == 1) {
 			Player player = args.getPlayer(0, false);
-			player.get(PlayerInventory.class).clear();
-			player.sendMessage(ChatStyle.RESET, plugin.getPrefix(), ChatStyle.BRIGHT_GREEN, "Your inventory has been cleared.", ChatStyle.RESET);
+			if (player == null) {
```
windy1, line 82: This will never be null `getPlayer()` will throw a CommandException if the player is not found.

src/main/java/org/spout/vanilla/command/AdministrationCommands.java
```diff
@@ -70,17 +70,30 @@ public void clear(CommandContext args, CommandSource source) throws CommandExcep
...
+				source.sendMessage(plugin.getPrefix(), ChatStyle.RED, "Player is not online!");
+				return;
+			}
+			PlayerInventory inv = player.get(PlayerInventory.class);
+			if (inv == null) {
+				source.sendMessage(plugin.getPrefix(), ChatStyle.RED, "Player has no inventory!");
```
windy1, line 88: `throw new CommandException(message)`

Zidane, line 88: Doesn't that print to the console? The exception spam isn't desired: if the player has no inventory cleanly say so and return.

zml2008, line 88: CommandException prints an error message to the player

Zidane, line 88: @zml2008 
We can't format that message :P or can we?

windy1, line 88: It just prints it in red. Personally I hate the convention of printing the plugins prefix on all the messages, it seems tacky and undesirable for users. Think about it, players don't give a shit what plugin is printing what message. If I'm the player I just want the information.

Zidane, line 88: I think quite different :P. I like to know what plugin is printing out what information.

PR #457, opened 2012-12-30
Implement basic pathfinding in Vanilla

fullwall:


src/main/java/org/spout/vanilla/component/living/Living.java
```diff
@@ -43,6 +45,9 @@ public void onAttached() {
 		holder.add(HealthComponent.class);
 		holder.add(PhysicsComponent.class);
 		holder.add(DrowningComponent.class);
+		holder.add(NavigationComponent.class);
+		holder.get(NavigationComponent.class).setDefaultExaminers(new VanillaBlockExaminer());
+
```
kitskub: You can just do holder.add(NavigationComponent.class).setDefaultExaminers...

fullwall: Thanks, will fix.

src/main/java/org/spout/vanilla/component/living/Living.java
```diff
@@ -43,6 +45,8 @@ public void onAttached() {
 		holder.add(HealthComponent.class);
 		holder.add(PhysicsComponent.class);
 		holder.add(DrowningComponent.class);
+		holder.add(NavigationComponent.class).setDefaultExaminers(new VanillaBlockExaminer());
+
```
kitskub, line 49: blank line?

PR #460, opened 2012-12-30
Event cause system

dredhorse:
For discussion only

https://github.com/SpoutDev/SpoutAPI/pull/486


dredhorse:
btw... the second commit is the one I'm talking about:   Example for new EventCauseSystem based on the BlockGrowCause / Event

PR #467, opened 2013-01-02
Refresh the client when broken blocks are cancelled via an interact event

Hidendra:
Hello again

Currently when you break a block a `PlayerInteractEvent` is thrown and if this is not cancelled a `BlockChangeEvent` will be thrown (signalling the block break). This assumes instant break. In the event of normal digging, `PlayerInteractEvent` is thrown when started and a `BlockChangeEvent` is thrown once it finally breaks. This is of course as expected.

However in both of these cases if the `PlayerInteractEvent` is cancelled, the player will most likely be left with a broken block in their client. The server should send the correct block back to the client upon failure of breaking the block.

This commit sends the block back to the client under 2 circumstances:
- In the `STATE_START_DIGGING` dig state, when the `PlayerInteractEvent` is cancelled **AND** the block is instantly breakable (creative or hardness = 0)
- In the `STATE_DONE_DIGGING` dig state, when the digging check fails **AND** the server sees the player as not currently digging (e.g the `PlayerInteractEvent` from `STATE_START_DIGGING` was cancelled!)

This has been confirmed to work as one would expect in both creative (instantly breaking blocks) and survival modes (digging).

### DIFF
--- a/src/main/java/org/spout/vanilla/protocol/handler/player/PlayerDiggingHandler.java
+++ b/src/main/java/org/spout/vanilla/protocol/handler/player/PlayerDiggingHandler.java
@@ -144,6 +144,9 @@ public final class PlayerDiggingHandler extends MessageHandler<PlayerDiggingMess
 		if (state == PlayerDiggingMessage.STATE_START_DIGGING) {
 			PlayerInteractEvent event = new PlayerInteractEvent(player, block.getPosition(), heldItem, Action.LEFT_CLICK, isInteractable, clickedFace);
 			if (Spout.getEngine().getEventManager().callEvent(event).isCancelled()) {
+				if (human.isCreative() || blockMaterial.getHardness() == 0.0f) {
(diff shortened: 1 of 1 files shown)


ghost:
Could you Signed-off-by this please?
http://wiki.spout.org/Pull_Request_Guidelines

Hidendra:
ah yeah, totally forgot on this one. Done now.

ghost:
btw, how about breaking signs? Will it update them :D?

Hidendra:
It does not currently, I could probably update the commit to include that?

ghost:
I think it would be cool feature to have, in Bukkit after breaking signs in protected area they become empty.

zml2008:
That's a Bukkit bug. Fixing it is not really a feature. It'll happen though.

Hidendra:
I created a separate branch and reverting sign text is working as well. It isn't really related to this PR directly so I will just send a second PR if this one is accepted. ([de097e24](https://github.com/Hidendra/Vanilla/commit/de097e24c7ea9cfd3d51ca145d538615ecabc97e))

PR #471, opened 2013-01-05
Happy New Year

Sleaker:


lukespragg:
This is already changed locally. Thanks for the though though. :-P
Generally, the pom and readme shouldn't be touched, same applies for formatting.

PR #474, opened 2013-01-06
Implemented basic Item dropping of mobs and animals.

timmhirsens:

### DIFF
--- a/src/main/java/org/spout/vanilla/component/living/hostile/Blaze.java
+++ b/src/main/java/org/spout/vanilla/component/living/hostile/Blaze.java
@@ -26,9 +26,14 @@
  */
 package org.spout.vanilla.component.living.hostile;
 
+import java.util.Random;
+
+import org.spout.api.inventory.ItemStack;
 import org.spout.vanilla.VanillaPlugin;
 import org.spout.vanilla.component.living.Hostile;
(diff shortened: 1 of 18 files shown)


src/main/java/org/spout/vanilla/component/living/hostile/Blaze.java
```diff
@@ -41,6 +46,9 @@
 	public void onAttached() {
 		super.onAttached();
 		getOwner().getNetwork().setEntityProtocol(VanillaPlugin.VANILLA_PROTOCOL_ID, new BlazeEntityProtocol());
+		getOwner().add(DropComponent.class);
+		Random random = new Random();
+		getOwner().get(DropComponent.class).addDrop(new ItemStack(VanillaMaterials.BLAZE_ROD, random.nextInt(1)));
```
Afforess: You keep repeating 'getOwner().get(DropComponent.class)' in onAttached. I suggest you store the DropComponent in a variable from the getOwner().add(...) call. That way you are not performing a bunch of useless hashmap lookups that get(...) performs.

Afforess:
Don't pull this until @fr1zle fixes the mentioned issue.

timmhirsens:
nvm, 1sec ;)

PR #475, opened 2013-01-06
Use getRandom() instead of creating one.

timmhirsens:

### DIFF
--- a/src/main/java/org/spout/vanilla/component/living/passive/Sheep.java
+++ b/src/main/java/org/spout/vanilla/component/living/passive/Sheep.java
@@ -122,7 +122,7 @@ public class Sheep extends Living implements Passive {
 	}
 
 	public boolean shouldEatGrass() {
-		final Random random = new Random();
+		final Random random = getRandom();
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/component/living/passive/Sheep.java
```diff
@@ -122,7 +122,7 @@ private void eatGrass() {
 	}
 
 	public boolean shouldEatGrass() {
-		final Random random = new Random();
+		final Random random = getRandom();
 		final Point position = getOwner().getTransform().getPosition();
```
Raphfrk, line 126: That is not a good way to get a Random.

It uses the world age as seed, which is worse than the default method.

I added a getRandom() method to MathHelper which uses a secure random seed a standard Random as a ThreadLocal variable.  

@bergerkiller has used it already.

You can access it with MathHelper.getRandom();  

It generates one per thread, so is thread safe.

Raphfrk, line 126: In fact, best plan would be to just change the method.

PR #500, opened 2013-01-16
Dye and Armor Textues ^^

Waterpicker:
Some has to do them.


windy1:
Did you test this? I doubt this works because the source provided in items are the source from the top right corner and in pixels.

PR #501, opened 2013-01-18
Breakdown of Vanilla gui to help with Customization

tenowg:
I wanted to create a way to allow full customization of the GUI in Vanilla, so i wrote this.

I am by far not a professional coder, and this is probably not something you guys would want to use, but figured I would give it a shot. Feel free to reject, criticize this, help me make it work better.

Thanks

### DIFF
--- a/src/main/java/org/spout/vanilla/plugin/component/player/hud/CrosshairWidget.java
+++ b/src/main/java/org/spout/vanilla/plugin/component/player/hud/CrosshairWidget.java
@@ -26,13 +26,4 @@
  */
 package org.spout.vanilla.plugin.component.player.hud;
 
-import org.spout.api.gui.Widget;
-
-/**
- *
- * @author Craig <tenowg at thedemgel.com>
- */
-public interface CrosshairWidget {
-    
-    public void init(Widget crosshair, float scale, float start_X);
-}
+public abstract class CrosshairWidget extends GUIWidget {}
--- a/src/main/java/org/spout/vanilla/plugin/component/player/hud/CrosshairWidget.java
+++ b/src/main/java/org/spout/vanilla/plugin/component/player/hud/CrosshairWidget.java
@@ -1,0 +1,38 @@
+/*
+ * This file is part of Vanilla.
+ *
+ * Copyright (c) 2011-2012, Spout LLC <http://www.spout.org/>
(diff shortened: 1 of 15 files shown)


src/main/java/org/spout/vanilla/plugin/component/player/hud/CrosshairWidget.java
```diff
@@ -0,0 +1,38 @@
...
+ * License and see <http://spout.in/licensev1> for the full license, including
+ * the MIT license.
+ */
+package org.spout.vanilla.plugin.component.player.hud;
+
+import org.spout.api.gui.Widget;
+
+/**
+ *
+ * @author Craig <tenowg at thedemgel.com>
```
kitskub: There shouldn't be any @authors

greatman:
Seems fine to me globally but personally won't pull due to Spout client not working for me. Can't test unfortunatly.

tenowg:
As far as my testing goes it works in Client... would like to add Hearts and Drowning as well so all widgets are together, etc..

src/main/java/org/spout/vanilla/plugin/component/player/hud/CrosshairWidget.java
tenowg: ahh... Netbeans inserts them automaticly, will remove is this goes anywhere if need be

greatman:
Go for it for drowning, I don't plan to touch that component soon.

Afforess:
Looks like some formatting issues. Tabs not spaces. Use the right kind of line-endings.

tenowg:
Please forgive me here... not used to this ;) all tabs should be spaces, as that is my default setting in Netbeans, and code looks to check out here... as far as line endings, as far as I know I have set this using the information provided in commit guildlines.

EDIT: damn if I didn't read that backwards.. :( I will change to tabs.. used to using default spaces for everything.

tenowg:
I will push this again, I just found some major changes I need to make to make this work as planned..

Afforess:
> all tabs should be spaces, as that is my default setting in Netbeans

See the readme on this repository: _Use tabs, no spaces._

http://wiki.spout.org/Pull_Request_Guidelines

PR #517, opened 2013-01-22
Switched from ticks to seconds for item spawn time

Grinch:


src/main/java/org/spout/vanilla/plugin/component/substance/Item.java
```diff
@@ -120,8 +120,9 @@ public static Item drop(Point position, ItemStack itemStack, Vector3 velocity) {
 
 	@Override
 	public void onTick(float dt) {
-		if (timeLeft-- <= 0) {
-			getOwner().remove();
+		timeLeft -= dt;
+			if (timeLeft <= 0) {
```
Zidane: Formatting

PR #518, opened 2013-01-22
Move more inventory events

Sleaker:


Afforess:
Why was this pulled? Empty classes/interfaces screams bad design.

Zidane:
@Afforess they are stubs for now :P

Afforess:
Then add them when you need them. They serve no purpose.

Sleaker:
@Afforess - I was told nothing in API can reference classes in the plugin package, to be able to move anything over to API (see events), these have to be housed in the API (and anything they reference), not in the plugin.  I don't see a way around that without having intermediate classes that serve as the API reference, or just moving everything around so it makes no difference which package it is in.

PR #520, opened 2013-01-22
move effects and effect events to API

Sleaker:


Sleaker:
Missed an import on the soundeffect.

PR #526, opened 2013-01-23
Add CommandBlocks, and custom wrapped packet handling

Snaipe:


Snaipe:
As a note, there is an issue with the first client update - clients won't see the input when they first connect/interact with the command block. I'm working on finding a good way to handle that, but if you have any idea, please let me know.

zml2008:
A couple major issues:
- We already have Packet250 support. Implementing it again just causes issues. I think there are a few other packets in Vanilla that already use the system, see them for examples. 
- Command blocks access the same commands that the rest of the server does. Just use the server's root command. 
- Some of your formatting is wrong.

Snaipe:
@zml2008 
I tried to reproduce the same structure as register, unregister, and ChunkCache which seem to be using Packet250, but it never worked... What should I do to have it working without doing what I did ?
I had two reasons to use an other root command : first, because commandBlocks does not use all of the commands the rest of the server does. /op, /deop, /kick, /ban, /stop, ... those commands are banned from the system. I don't think that putting the allowed/banned commands in a config is a good solution either, so making a separate root command would make a simple and clear alternative. The second reason is that if I use the server root command, all adventure maps using command blocks would break. The command formats are different from the vanilla ones ( namely, /tp <player> <x> <y> <z>, /give, ... ).
Regarding the wrong formatting, I missed that, where is it wrong ?

src/main/java/org/spout/vanilla/plugin/component/substance/material/CommandBlockComponent.java
```diff
@@ -0,0 +1,631 @@
...
+		return false;
+	}
+
+	@Override
+	public Locale getPreferredLocale() {
+		return Locale.ENGLISH_US;
+	}
+	
+	public void executeCommandInput()
+	{
```
zml2008, line 131: brackets here are wrong

src/main/java/org/spout/vanilla/plugin/protocol/VanillaProtocol.java
```diff
@@ -68,7 +70,7 @@
 	public final static DefaultedKey<Long> LOGIN_TIME = new DefaultedKeyImpl<Long>("handshake_time", -1L);
 	public final static DefaultedKey<ChunkNetCache> CHUNK_NET_CACHE = new DefaultedKeyImpl<ChunkNetCache>("chunk_net_cache", (ChunkNetCache) null);
 	public static final DefaultedKey<ArrayList<String>> REGISTERED_CUSTOM_PACKETS = new DefaultedKey<ArrayList<String>>() {
-		private final List<String> defaultRestricted = Arrays.asList("REGISTER", "UNREGISTER");
+		private final List<String> defaultRestricted = Arrays.asList("REGISTER", "UNREGISTER", "MC|AdvCdm");
```
zml2008, line 73: this doesn't need to be modified.

zml2008:
Ah, it looks like somebody broke the dynamic-id message receiving. I fixed it in the most recent commit.

zml2008:
And with regards to the commands: That's why permissions exist...
For commands that have better syntax, just make the normal Vanilla commands support the MC syntax too.

windy1:
The command block should probably just inherit the permissions of who programmed it.

zml2008:
That doesn't really work :/ - The person who set the block might no longer
be online, and everybody was against offline players.

windy1:
@zml2008 Oh, I just assumed we had OfflinePlayers, why the opposition?

zml2008:
Nobody saw a purpose & they were too complicated.

Snaipe:
@zml2008 
I fixed the packet part and the formatting.
The only issue I see not to have a separate root now is that I've seen some commands with special processing for the command block (mostly because the output is broadcasted) - though that could be fixed with an source instanceof CommandBlockComponent; and that if a plugin overrides/replaces one of vanilla's commands, it will surely break the command blocks.
That's what bukkit did, and now some of the command block commands does not work properly/at all because of command providing plugins such as CommandBook (I often need to put maps on a separate vanilla server for my events because of that, and it sucks.)

Raphfrk:
> @zml2008 Oh, I just assumed we had OfflinePlayers, why the opposition?

The issue was on login, you could end up with 2 instances for the same player.

It might be worth having a specific offline player object for things like permission checking.  It would probably be read-only though.

windy1:
@Raphfrk it's not really a big deal, we can check if a player has vanilla.commandblock.<command> when they actually program the block or something.

Snaipe:
I had a little issue with git, I had to resend an other PR for that https://github.com/VanillaDev/Vanilla/pull/527

dredhorse:
why not store the permission into the command block when it is created? That would also allow to create a command block with more permissions and than afterwards demote yourself to normal player. To cover adventure maps you would use as default the server permissions.

PR #529, opened 2013-01-25
Lets use our MathHelper for all these Randoms, attempt #2.

Dockter:
Signed-off-by: Mike Howe dockter@almuramc.com

### DIFF
--- a/src/main/java/org/spout/vanilla/plugin/component/misc/HungerComponent.java
+++ b/src/main/java/org/spout/vanilla/plugin/component/misc/HungerComponent.java
@@ -40,6 +40,7 @@ import org.spout.api.geo.discrete.Point;
 import org.spout.api.gui.Widget;
 import org.spout.api.gui.component.RenderPartsHolderComponent;
 import org.spout.api.gui.render.RenderPart;
+import org.spout.api.math.MathHelper;
 import org.spout.api.math.Rectangle;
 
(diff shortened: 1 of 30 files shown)


src/main/java/org/spout/vanilla/plugin/component/misc/HungerComponent.java
```diff
@@ -71,7 +72,7 @@
 	// Client
 	private final Widget widget = new Widget();
 	private static final float SCALE = 0.75f; // TODO: Apply directly from engine
-	private final Random random = new Random();
+	private static final Random random = MathHelper.getRandom();
```
Raphfrk, line 75: No this is wrong.

MathHelper.getRandom() gets a Random for the current thread.

That way, you are guaranteed thread safety without having to create loads of Random instances.

However, it only works if you call the method when using the random

In this case, if the hunger component was created in one thread and one of the methods that uses the Random is called from multiple threads, then the same Random object would be accessed from multiple threads.

Afforess, line 75: @Raphfrk - good point, I also caught quite a few times outside of this specific pull request where Randoms were being cached inside class and static level variables. I have corrected all of the instances of use as far as I am aware:

https://github.com/VanillaDev/Vanilla/commit/950c44502bed90acbc046b57fa7b47f508bc1db1

PR #530, opened 2013-01-26
Inital work at fixing inventory rendering

Waterpicker:


src/main/java/org/spout/vanilla/plugin/inventory/util/GridInventoryConverter.java
```diff
@@ -55,7 +55,9 @@ public GridInventoryConverter(Inventory inventory, int length, int offset, Vecto
 			int i = iter.next(), x = iter.getX(), y = iter.getY(), size = grid.getSize();
 			slots[i] = (offset + size) - (length * y) - (length - x);
 			InventorySlot slot = widgets[i].get(InventorySlot.class);
-			slot.setRenderItemStack(new RenderItemStack(new ItemStack(VanillaMaterials.LEATHER_CAP, 1))); // TODO: Handle this in onClick
+			if(inventory.get(i) == null)
```
windy1, line 58: This is wrong. If _one_ ItemStack is null then the converter will stop placing slots altogether. Also, we can probably change it so RenderItemStacks are all handled internally and you shouldn't need to set them. You should only need to set the slot and the position.

Waterpicker, line 58: Windwaker. You are mistaken. This prevents exactly what you stated and I said inital

kitskub, line 58: Actually, no. It should be continue; not return;

windy1, line 58: ^ This, return will exit the loop entirely.

PR #535, opened 2013-01-27
Stopped resetting of gamemode during login.

Grinch:
The way it was prior to this was basically always take the World's gamemode specified in worlds.yml and apply it to the player when they log in. This change makes it so if its null, apply the world's default gamemode but if they have one specified VanillaData.GAMEMODE (Creative, Survival, Adventure) then to use that instead.

### DIFF
--- a/src/main/java/org/spout/vanilla/plugin/protocol/VanillaNetworkSynchronizer.java
+++ b/src/main/java/org/spout/vanilla/plugin/protocol/VanillaNetworkSynchronizer.java
@@ -422,6 +422,7 @@ public class VanillaNetworkSynchronizer extends NetworkSynchronizer implements P
 
 	@Override
 	protected void worldChanged(World world) {
+		GameMode gamemode;
 		WorldConfigurationNode node = VanillaConfiguration.WORLDS.get(world);
 		maxY = node.MAX_Y.getInt() & (~Chunk.BLOCKS.MASK);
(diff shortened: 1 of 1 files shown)


Afforess:
I don't see how this will fix anything

src/main/java/org/spout/vanilla/plugin/protocol/VanillaNetworkSynchronizer.java
```diff
@@ -430,7 +431,11 @@ protected void worldChanged(World world) {
 		highY = minY + stepY;
 		lastY = Integer.MAX_VALUE;
 
-		GameMode gamemode = world.getComponentHolder().getData().get(VanillaData.GAMEMODE);
+		if (player.getData().get(VanillaData.GAMEMODE) == null) {
```
Afforess: The game mode will never be null. The point of the defaulted key system (VanillaData.GAMEMODE) is that you can assign default values. So this branch will never be executed.

The default value for gamemode is survival: https://github.com/VanillaDev/Vanilla/blob/master/src/main/java/org/spout/vanilla/plugin/data/VanillaData.java#L64

Zidane: Lawl Grinch derped Affy

He was supposed to lookup the STRING as I told him. If get("gamemode") is null then that means they never had the data assigned to the Player and as such, assume the world's value.

PR #540, opened 2013-01-28
Inventory API & EntityInventories

Sleaker:
refactor how inventories are constructed to provide an API.
Adds inventory components to Zombie, Pig Zombie, Skeleton.
now have a 'Quickbar' for their held item slot, and ArmorInventory for armor.
CreatureProtocol now checks to see if an entity has Inventory and attempts to send worn/held item to clients.


greatman:
Basicly if I just do a command to add items to mobs I should see it wearing it right? (Just to test if it really works :P)

Sleaker:
I don't know if there's a command to add items onto them at the moment, but no. you wont necessarily see them unless you spawn them with the items at this point.  I don't think there's any update protocol linked up yet for sending worn item updates except for players.

PR #541, opened 2013-01-30
Add canEquip to Materials

Sleaker:
No longer use Inventory to control if a material is wearable in a specific inventory slot.
All materials default to not being equipable, now override canEquip.
Fixes Skulls & Pumpkins not being equipable in Helm slot.


Sleaker:
Will be re-writing this as discussed with @Zidane

PR #551, opened 2013-02-01
Add InventoryCanSetEvent

Sleaker:


windy1:
What about an InventorySetEvent in SpoutAPI?

Sleaker:
there's already inventory event, and SetEvent doesn't make much sense. inventory restrictions are a Vanilla thing, and I'm not sure there needs to be an inventorySet event in Spout, unless it's only function is to be extended.

See PlayerHeldItemChangeEvent which already exists.

Edit: I guess a SetEvent might be useful for the other cases there in the window.

PR #553, opened 2013-02-02
Move WindowEvents to API packages

Sleaker:
Created AbstractWindow.  VanillaRenderMaterials is also now in the API package as it's a static class and seems to be universally necessary.

Updated Docs for InventoryCanSetevent.  removed PlayerSlotChangeEvent, it was unnecessary to begin with.


src/main/java/org/spout/vanilla/api/inventory/window/AbstractWindow.java
```diff
@@ -0,0 +1,314 @@
...
+import org.spout.api.inventory.InventoryViewer;
+import org.spout.api.inventory.ItemStack;
+import org.spout.api.plugin.Platform;
+import org.spout.api.protocol.event.ProtocolEvent;
+
+import org.spout.vanilla.api.component.inventory.PlayerInventoryComponent;
+import org.spout.vanilla.api.inventory.Slot;
+import org.spout.vanilla.api.inventory.window.prop.WindowProperty;
+
+public abstract class AbstractWindow implements InventoryViewer {
```
windy1, line 49: Ehhhh.... I don't like this rename :P

Sleaker, line 49: got anything better?

windy1, line 49: nope, heh, I'll live

PR #563, opened 2013-02-04
Added slab and reversed slab obj files

SelennLamson:
in resources/materials/block/nonsolid/slab/..
As : kind(top)slab.obj

To link in the code.

Signed-off-by: Thomas LAMSON tomlabete@numericable.fr


Sleaker:
why are slabs different? isn't the model identical for all of the different slab types?

To prevent unnecessary additions shouldn't this just be a unified slab? the textures are the only thing different here right?

SelennLamson:
Just two slabs models in fact, top slab and bot slab.
But for the texture, unless it's improved in the source code, I'll make all obj files for each model where it's needed.
There is a real difference in the file, on the point's coordinates on the texture pack. (and I have a little soft too quickly change textures of my objs)

Sleaker:
@Tomlabete - why is there a different slab for each material type? I was asking, don't they all use identical model data and the only difference is the texture?

I thought the model was seperate from the texture.

windy1:
Yeah, this can be limited to two meshes couldn't it? You just need to UV map the mesh in away that works for each material.

DDoS:
Spout doesn't support reskinning models at the moment.

windy1:
@DDoSQc You would just need to make multiple spm files, you can still stick to one mesh. For example you can do this with one mesh:

``` yaml
Mesh: mesh://Vanilla/path/to/mesh
Material: material://Vanilla/path/to/material
```

and keep the same mesh but use a different texture for each one.

DDoS:
The .OBJ file needs to define the UV coords.

Sleaker:
@DDoSQc why do the UV coordinates need to be different if you're just rendering a flat face on each side with the exact same dimensions?  regardless of the material, all slabs have a side texture and a top/bottom texture, no?

windy1:
@DDoSQc Right, but UV coords to not have to be exclusive to textures.

DDoS:
The UV coords the .OBJ has are mapped to terrain.png. That's the issue right now.

windy1:
Ah right, well that is changing in 1.5, every texture is getting it's own file.

Sleaker:
I would suggest developing the textures the way the way we want the spout client to handle them, not the way MC handles them necessarily.  since MC is moving to not having a shared terrain.png, this commit becomes largely obselete in a couple weeks.

windy1:
Right, there's no point in adding to the old system since it's going to become outdated very soon. Closing this now because it doesn't make too much sense to pull it if it will have to be replaced soon anyways.

PR #564, opened 2013-02-05
Monsters equipping items

Sleaker:
Added new MonsterPickupitemComponent to handle logic for hostiles picking up inventory items they can wear.  Zombies, Skeletons, and PigZombies should now pickup items they walk over if they are better than items they currently have equipped, they will also pickup items into their held slot if nothing is equipped.

### DIFF
--- a/src/main/java/org/spout/vanilla/plugin/VanillaListener.java
+++ b/src/main/java/org/spout/vanilla/plugin/VanillaListener.java
@@ -53,7 +53,7 @@ import org.spout.vanilla.plugin.component.living.neutral.Human;
 import org.spout.vanilla.plugin.component.misc.HealthComponent;
 import org.spout.vanilla.plugin.component.misc.HungerComponent;
 import org.spout.vanilla.plugin.component.misc.LevelComponent;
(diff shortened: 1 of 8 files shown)


Sleaker:
I'm not sure if it's correct to call the protocol event for the entity picking up the item. If this is incorrect please let me know.

windy1:
Are inventories working for you atm? I've been having some issues with them.

Sleaker:
@W1ndwaker - The Widget refactor broke changes being sent to players. Grinch tested it out pretty thoroughly, and they were working up until that commit.  So something in that refactor broke the client being sent updates.  He noticed that items are added/set server-side, but they aren't sent as updates to the player.  I threw a message to Karang since it looked like he had done a bit of the refactor and I'm really lost as to why a Widget refactor would break inventory changes being sent to the client.

Afforess:
I can verify part of Sleaker's assertions, I have gone into debug mode and inventories themselves work (they collect, retain items and state correctly), but the changes are not being sent.

Zidane:
@Sleaker is changing a few things for me, don't pull.

PR #565, opened 2013-02-05
Added enchantment mechanics on enchantment table

stewbasic:
*Added methods to api/enchantment/Enchantment.java to generate and add random enchantments
*Logic based on Minecraft 1.4.7
*Changed "enchantment level" to "power level" in places to distinguish between the power level of an enchantment (I, II, III, IV, V) and the enchantment level displayed in the enchantment interface (1 to 30)
*Fixed bug in Enchantments.values(), which was assuming the enchantment ids were consecutive
*Fixed bug in Enchantment.addEnchantment etc, which weren't putting the NBT data in the correct format (http://www.minecraftwiki.net/wiki/Player.dat_Format#Item_structure)
*Fixed bug in ToolEnchantment.canEnchant (returned true on bows and swords)
*Simplified xp code
*Update EnchantmentTable's level display when it changes, rather than every tick
*Allow bookshelves on one side of table to contribute despite blocks on the other side (still not quite identical to Vanilla)
*Moved enchantment levels from EnchantmentTableInventory to EnchantmentTableWindow
*Fixed bug in Window.handleClick, event should be cancelled if the slot cannot be set
*Added tests EnchantmentTest and LevelComponentTest
Signed-off-by: stewbasic stewbasic@gmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/plugin/component/substance/material/EnchantmentTable.java
+++ b/src/main/java/org/spout/vanilla/plugin/component/substance/material/EnchantmentTable.java
@@ -52,19 +52,6 @@ public class EnchantmentTable extends EnchantmentTableComponent {
 	}
 
 	@Override
-	public void onTick(float dt) {
-		EnchantmentTableInventory inventory = getInventory();
-		// Only display enchantment levels if there is an item and it is not enchanted
-		if (inventory.has()) {
-			for (Player player : viewers) {
-				for (int i = 0; i < 3; i++) {
-					player.get(WindowHolder.class).getActiveWindow().setProperty(i, !Enchantment.isEnchanted(inventory.get()) ? inventory.getEnchantmentLevel(i) : 0);
-				}
-			}
-		}
-	}
-
-	@Override
 	public boolean open(Player player) {
 		EnchantmentTableOpenEvent event = Spout.getEventManager().callEvent(new EnchantmentTableOpenEvent(this, player));
 		if (!event.isCancelled()) {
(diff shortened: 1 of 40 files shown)


src/main/java/org/spout/vanilla/plugin/component/substance/material/EnchantmentTable.java
```diff
@@ -93,33 +80,27 @@ public int getNearbyBookshelves() {
...
-						continue; // Ignore the enchantment table itself
+		// if there are any blocks right next to the enchantment table, bookshelves in that direction are nullified
+		// Note: unlike Mojang's version, a block at (x+1,y+1,z) won't nullify a bookshelf at (x+2,y,z)
+		int bookshelves = 0;
+		for (int dx = -1; dx <= 1; dx++) {
+			for (int dz = -1; dz <= 1; dz++) {
+				if (dx == 0 && dz == 0)
+					continue; // Ignore the enchantment table itself
+				for (int yy = y; yy <= y + 1; yy++) {
+					if (!VanillaMaterials.AIR.equals(block.getWorld().getBlock(x + dx, yy, z + dz)))
```
Afforess: Minor note: block.translate(dx, yy, dz) is much more efficient than block.getWorld().getBlock(). The block class has optimizations that allow avoiding the block->world->region->chunk->block lookup and instead goes straight from block -> chunk -> block instead, avoiding hashmap lookups and only uses array access.

This is a note for project maintainers, it should not stop the PR from being pulled.

Sleaker:
This needs to have all uses of .has(SomeComponent.class) removed and instead use .get(SomeComponent.class) instead, this change may be as simple as rebasing, as I've already removed all uses of these method calls in Vanilla.  At the moment this PR is not mergable due to changes.

stewbasic:
I added a merged commit. Should I close and reopen the pull request?

Sleaker:
@stewbasic - you should rebase + squash your commits into one.

stewbasic:
After rebasing, the commit's only parent is master, so I can't push without using -f, presumably removing my previous commits and this pull request. Is this right?

windy1:
You will have to `git push --force` in order to squash your already pushed commits.

stewbasic:
OK, done. Thanks for explaining!

greatman:
Need another rebase. Can't be merged.

stewbasic:
Rebased onto Sleaker's exp commit.

greatman:
Ill test this tonight when i'm home. It's my duty to work on enchants & Potions area of Vanilla.

greatman:
So. Atm in my testing I crash when I enchant a diamond sword and close the window.

Also, XP doesn't show on login but I can fix that easily.

EDIT: Also, I can't seem to get a Enchant higher than lvl 7.

The bug I list are maybe already existing in current vanilla. Didin't verify.

stewbasic:
Hm, I don't experience either of these issues. I'm not sure what could cause the different behaviour. I know the crash does occur on the current master branch.

Anyway I'll do another rebase since master has some more changes (though it's a clean merge this time).

PR #569, opened 2013-02-07
Convenience Methods for Minecraft Protocol and distinction in CreatureType

ayyron-dev:
This PR adds some convenience methods for CreatureType adding Protocol Names and allowing the CreatureType to be gotten from the Entity.

https://github.com/somners/Vanilla/blob/87a4fe35a224c2b74a0491706ace68c6a638873f/src/main/java/org/spout/vanilla/plugin/protocol/entity/creature/CreatureType.java


greatman:
Needs to be squashed into 1 commit & signed.

Zidane:
Don't pull this Vanilla Team. We are going to try another route that broadens the scope a bit.

Sleaker:
you can already get the creaturetype of an entity by passing it into the creaturetype class, eg: CreatureType.get(entity.get(Living.class)).  if you need a name of the creaturetype you can toString() it.

PR #572, opened 2013-02-09
Fix for: Group Renderparts by material to speed up the gui render

Karang:


src/main/java/org/spout/vanilla/plugin/inventory/window/gui/RenderItemStack.java
```diff
@@ -49,7 +47,6 @@ public RenderItemStack(ItemStack item) {
 			throw new IllegalArgumentException("ItemStack must have a VanillaMaterial for a Material.");
 		}
 		this.item = item;
-		setRenderMaterial(VanillaRenderMaterials.ITEMS_MATERIAL);
```
windy1, line 52: Y U Remove this?

PR #589, opened 2013-02-14
Add client chat handler, comment some NPE

Karang:


src/main/java/org/spout/vanilla/plugin/protocol/handler/player/PlayerChatHandler.java
```diff
@@ -35,8 +37,19 @@
...
+	
 	@Override
-	public void handle(boolean upstream, Session session, PlayerChatMessage message) {
+	public void handleClient(Session session, PlayerChatMessage message) {
+		if (!session.hasPlayer()) {
+			return;
+		}
+
+		ChatArguments text = ChatArguments.fromString(message.getMessage(), VanillaStyleHandler.ID);
+		((Client)Spout.getEngine()).getScreenStack().getConsole().addMessage(text);
```
zml2008, line 48: You should be able to do `((Client) session.getEngine()).getCommandSource().sendMessage(text);`

src/main/java/org/spout/vanilla/plugin/protocol/handler/entity/pos/EntityTeleportHandler.java
```diff
@@ -45,7 +45,7 @@ public void handleClient(Session session, EntityTeleportMessage message) {
 		
 		Player player = session.getPlayer();
 		World world = player.getWorld();
-		Entity entity = world.getEntity(message.getEntityId());
+		Entity entity = session.getPlayer();//world.getEntity(message.getEntityId());
```
zml2008, line 48: I think this packet can refer to other entities

Karang, line 48: true but we need to add a way to set an id to an entity because this is getting a NPE
I have told @L5D, him or i willl continue this tomorrow

zml2008, line 48: yeah, that's needed for the Spout protocol too. Not entirely sure how I
want that to work, but I've been experimenting with it a bit. If you come
across a good solution, feel free to implement it in Spout.

PR #594, opened 2013-02-26
Fixes Chiseled Stone Brick texture when placed.

ayyron-dev:
When placing a Chiseled Stone Brick the texture was viewed as a regular Stone Brick. This PR fixes that.


windy1:
Formatting is off, use tabs not spaces.

ayyron-dev:
I was just told that Netbeans changes tabs to spaces. So that is why that is like that. I'll fix it.

lukespragg:
The commit also needs to be signed off. See http://spout.in/prguide

lukespragg:
Did you mean to close this? :P

PR #597, opened 2013-02-27
Water and death stop burning

olliestanley:


natemort:
Why hasn't this been pulled?

src/main/java/org/spout/vanilla/component/entity/misc/Burn.java
```diff
@@ -64,10 +64,12 @@ public void onTick(float dt) {
 			} else {
 				rainTimer = 0f;
 			}
-			if (rainTimer >= 2.0f) {
+			if (rainTimer >= 2.0f || point.getBlock().getMaterial() == VanillaMaterials.WATER || health.isDead()) {
```
Afforess, line 67: == Water is bad. It doesn't cover stationary water. Use instanceof Water instead.

Afforess:
Probably because there are problems with it.

src/main/java/org/spout/vanilla/component/entity/misc/Burn.java
olliestanley, line 67: Shall do, it should be done today hopefully

Sleaker:
re-implemented a bit different, thanks @DziNeIT for the attempt.

PR #599, opened 2013-03-08
Fix to InputCommands. Needs additional safety fix.

trentonfaris:
Signed-off-by: Man_cub mcmancub@gmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/command/InputCommands.java
+++ b/src/main/java/org/spout/vanilla/command/InputCommands.java
@@ -46,11 +46,13 @@ import org.spout.vanilla.inventory.window.Window;
 import org.spout.vanilla.material.VanillaMaterials;
 
 public class InputCommands {
+	private final VanillaPlugin plugin;
 	private final Client client;
 	private BlockMaterial selection;
 
-	public InputCommands(Client client) {
-		this.client = client;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spout/vanilla/command/InputCommands.java
```diff
@@ -46,11 +46,13 @@
 import org.spout.vanilla.material.VanillaMaterials;
 
 public class InputCommands {
+	private final VanillaPlugin plugin;
 	private final Client client;
 	private BlockMaterial selection;
 
-	public InputCommands(Client client) {
-		this.client = client;
+	public InputCommands(VanillaPlugin plugin) {
```
windy1, line 53: This doesn't change anything and it's not unsafe because this is only created in VanillaPlugin when the Engine is in client mode.

trentonfaris, line 53: That's what I had observed. I just figured in a general sense not adding a check for it would be considered unsafe. But you do whatever with it :)

MC

windy1:
@MCManCub I have closed this because it does not cause an issue, the error you showed me is due to an outdated API on your end.

EDIT: My mistake, this does fix the error, but not for the reason you think it does. This fixes that the passed param needs to be a VanillaPlugin, not a casting issue. My mistake.

PR #601, opened 2013-03-16
Implement animals following Players when holding food.

timmhirsens:

### DIFF
--- a/src/main/java/org/spout/vanilla/component/entity/living/passive/Chicken.java
+++ b/src/main/java/org/spout/vanilla/component/entity/living/passive/Chicken.java
@@ -30,10 +30,12 @@ import org.spout.api.geo.discrete.Point;
 import org.spout.api.inventory.ItemStack;
 import org.spout.api.math.GenericMath;
 import org.spout.api.math.Vector3;
-
 import org.spout.vanilla.VanillaPlugin;
-import org.spout.vanilla.component.entity.living.Passive;
+import org.spout.vanilla.ai.action.FollowMaterialHolderAction;
+import org.spout.vanilla.ai.goal.FollowMaterialHolderGoal;
+import org.spout.vanilla.ai.sensor.NearbyMaterialHolderSensor;
 import org.spout.vanilla.component.entity.living.Animal;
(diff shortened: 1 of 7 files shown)


src/main/java/org/spout/vanilla/component/entity/living/passive/Chicken.java
```diff
@@ -63,6 +65,11 @@ public void onAttached() {
 		if (getAttachedCount() == 1) {
 			getOwner().add(Health.class).setSpawnHealth(4);
 		}
+		NearbyMaterialHolderSensor materialHolderSensor = new NearbyMaterialHolderSensor(getAI(), VanillaMaterials.SEEDS);
```
windy1, line 68: Couldn't this be abstracted to Animal instead of putting it in every Animal extension?

timmhirsens, line 68: Probably, yeah.
Maybe that would make it harder for plugins to implement animals that DON'T follow the player, though? 

Edit: And there are animals in Vanilla that don't follow the player if he is holding a specific food item.

windy1, line 68: No it would be easy to remove the AI, what Entities that extends Animal don't follow the player? e: Ah I see, some animals follow different animals.

timmhirsens, line 68: Ocelots and Wolves.

windy1, line 68: Wolves will follow any kind of meat and ocelots follow fish: http://www.minecraftwiki.net/wiki/Breeding

timmhirsens, line 68: I tried it ingame (official) and they did not follow me. The Items can be used to tame and breed them.
Once they are tamed they will always follow the player. Thats a different kind of AI though.

timmhirsens:
Is there anything that needs to be done before this can be pulled?

PR #602, opened 2013-03-17
Removal of Spout Class Referecneces Attempt 2.

Waterpicker:
...nt side works

Waterpicker Waterpicker@rocketmail.com

### DIFF
--- a/src/main/java/org/spout/vanilla/VanillaListener.java
+++ b/src/main/java/org/spout/vanilla/VanillaListener.java
@@ -28,7 +28,6 @@ package org.spout.vanilla;
 
 import org.spout.api.Client;
 import org.spout.api.Platform;
-import org.spout.api.Spout;
 import org.spout.api.component.impl.CameraComponent;
 import org.spout.api.component.impl.InteractComponent;
 import org.spout.api.entity.Player;
@@ -102,11 +101,11 @@ public class VanillaListener implements Listener {
 
 	@EventHandler
 	public void onGameStart(EngineStartEvent event) {
-		if (Spout.getPlatform() != Platform.CLIENT) {
+		if (VanillaPlugin.getInstance().getEngine().getPlatform() != Platform.CLIENT) {
 			return;
 		}
 
-		Player player = ((Client) Spout.getEngine()).getActivePlayer();
+		Player player = ((Client) VanillaPlugin.getInstance().getEngine()).getActivePlayer();
 
 		HUD HUD = player.add(org.spout.vanilla.component.entity.player.HUD.class);
 		HUD.setDefault(VanillaArmorWidget.class);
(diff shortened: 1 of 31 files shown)


Sleaker:
I don't like just blindly passing around the VanillaPlugin instance all over the place. Also not a a fan of the static access for the filesystem reference, depending on how classloading works it's liable to produce unknown behaviour (ie break)

I think a few of these classes need to be changed so they aren't just static lists, they need to be instantiated, or passed in a reference to the engine (recipe manager etc).  

I'm not sure if this can be pulled, though it looks like it's mostly fine.

Waterpicker:
Yea. The purpose of this pull is not meant to be a final edit. Just enough so Affy can go ahead with his work on launching and stuff.

src/main/java/org/spout/vanilla/VanillaListener.java
```diff
@@ -102,11 +101,11 @@ public void onPlayerJoin(PlayerJoinEvent event) {
 
 	@EventHandler
 	public void onGameStart(EngineStartEvent event) {
-		if (Spout.getPlatform() != Platform.CLIENT) {
+		if (VanillaPlugin.getInstance().getEngine().getPlatform() != Platform.CLIENT) {
```
Afforess, line 104: Event should contain engine instance

src/main/java/org/spout/vanilla/VanillaListener.java
```diff
@@ -164,7 +163,7 @@ public void onBlockChange(BlockChangeEvent event) {
 		}
 		if (prevPower != -1) {
 			RedstoneChangeEvent redstoneEvent = new RedstoneChangeEvent(event.getBlock(), event.getCause(), prevPower, newPower);
-			Spout.getEventManager().callEvent(redstoneEvent);
+			VanillaPlugin.getInstance().getEngine().getEventManager().callEvent(redstoneEvent);
```
Afforess, line 166: Doesn't the listener get passed the engine instance when it is created? If not, it should.

src/main/java/org/spout/vanilla/component/entity/substance/test/SceneOutput.java
```diff
@@ -26,15 +26,16 @@
...
-import org.spout.api.Spout;
+
 import org.spout.api.component.type.EntityComponent;
 import org.spout.api.entity.Player;
 
 public class SceneOutput extends EntityComponent {
 	@Override
 	public void onAttached() {
-		if (Spout.getPlatform() != Platform.SERVER || !Spout.debugMode() || getOwner() instanceof Player) {
+		if (getOwner().getEngine().getPlatform() != Platform.SERVER || !getOwner().getEngine().debugMode() || getOwner() instanceof Player) {
```
Afforess, line 38: No need for getOwner().getEngine(), you can do just getEngine()

windy1:
Do we actually _need_ `VanillaPlugin.getInstance()` for anything besides convenience? If not, I think we should remove it anyways.

Waterpicker:
I only used what I had at my disposal.

Zidane:
@W1ndwaker 

Sadly Materials need it for initialize in VanillaMaterials else we would have already removed it.

Afforess:
Merged.

PR #603, opened 2013-03-21
Lever models

Meiun:
everything should be correct. Up & Down are just rotations of N and S
respectively.

Signed-off-by: Hung-Lin nymphalis.aloft@gmail.com


Meiun:
in E/W & N/S the central shaft is rotated 180. 
in E/W Vs N/S The textures on top of the block is rotated.

PR #604, opened 2013-03-23
Adds last damager and damage cause to VanillaEntityDeathEvent and PlayerDeathEvent

timmhirsens:

### DIFF
--- a/src/main/java/org/spout/vanilla/component/entity/misc/Health.java
+++ b/src/main/java/org/spout/vanilla/component/entity/misc/Health.java
@@ -215,9 +215,9 @@ public class Health extends EntityComponent {
 		VanillaEntityDeathEvent event;
 		Entity owner = getOwner();
 		if (owner instanceof Player) {
-			event = new PlayerDeathEvent((Player) owner);
+			event = new PlayerDeathEvent((Player) owner, lastDamageCause, lastDamager);
 		} else {
-			event = new VanillaEntityDeathEvent(owner);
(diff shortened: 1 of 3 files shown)


Sleaker:
lastDamager needs to be more type-specific than an object. having it so unspecific doesn't make the API friendly to plugins, this may have been why nothing was implemented yet.

dredhorse:
Problem is I think that there are a lot of different damagers... minecarts, cactus, lava, player, mobs etc..

I don't really know a good solution to this though myself, except perhaps having lastDamageCause handle it, together with an enum which exposes the damager. On the other side that makes it problematic to extend.

windy1:
@Sleaker @dredhorse is right, I don't see any nicer way to do it, no matter how you implement it you will always have to cast something.

Sleaker:
@W1ndwaker - then putting the Damager on the Cause is the best way to handle this, rather than making it accessible directly.

Providing a method that returns an Object does no good as there's no functional way to tell what it is without performing a considerable number of checks, and even then you can't be certain what you're getting out.

timmhirsens:
It is the same in EntityDamageEvent. This is of course not a good solution but I really wouldn't know how to handle it different / better.
Only exposing it in DamageCause doesn't really change anything imho, since its just one method call more.

windy1:
@Sleaker 

What's so complicated about.

``` java
Object lastDamager = event.getLastDamager();
if (lastDamager instanceof Player) {
    Player player = (Player) lastDamager;
    // last damager is a player
}
```

Do you have a simpler solution?

PR #609, opened 2013-05-14
Blocks can now have multiple components

kitskub:
Discussion can be found here: http://forums.spout.org/threads/block-components.6376/

Spout: https://github.com/SpoutDev/Spout/pull/2196
SpoutAPI: https://github.com/SpoutDev/SpoutAPI/pull/550

### DIFF
--- a/src/main/java/org/spout/vanilla/material/block/redstone/RedstoneComparator.java
+++ b/src/main/java/org/spout/vanilla/material/block/redstone/RedstoneComparator.java
@@ -100,6 +100,7 @@ public class RedstoneComparator extends GroundAttachable implements Directional,
 	public void onPlacement(Block block, short data, BlockFace against, Vector3 clickedPos, boolean isClickedBlock, Cause<?> cause) {
 		super.onPlacement(block, data, against, clickedPos, isClickedBlock, cause);
 		this.setFacing(block, PlayerUtil.getFacing(cause));
+        /*//TODO this is broken; nothing implements Container; there's no way to get an interface from a componentholder
 		Block input = block.translate(this.getFacing(block).getOpposite());
 		if (input.getMaterial().isMaterial(inputs)) {
 			Inventory inventory = ((Container) input.getComponent()).getInventory();
@@ -114,11 +115,12 @@ public class RedstoneComparator extends GroundAttachable implements Directional,
 			if (!viewed) {
 				inventory.addViewer(new ComparableViewer(block, inventory));
 			}
-		}
+		}*/
 	}
 
 	@Override
 	public boolean onDestroy(Block block, Cause<?> cause) {
+        /*//TODO this is broken; nothing implements Container; there's no way to get an interface from a componentholder
 		Block input = block.translate(this.getFacing(block).getOpposite());
 		if (input.getMaterial().isMaterial(inputs)) {
 			Inventory inventory = ((Container) input.getComponent()).getInventory();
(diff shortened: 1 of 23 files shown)


Afforess:
Pulled, thanks.

src/main/java/org/spout/vanilla/material/block/redstone/RedstoneComparator.java
```diff
@@ -100,6 +100,7 @@ public void setPowered(Block block, boolean powered) {
 	public void onPlacement(Block block, short data, BlockFace against, Vector3 clickedPos, boolean isClickedBlock, Cause<?> cause) {
 		super.onPlacement(block, data, against, clickedPos, isClickedBlock, cause);
 		this.setFacing(block, PlayerUtil.getFacing(cause));
+        /*//TODO this is broken; nothing implements Container; there's no way to get an interface from a componentholder
```
windy1, line 103: @Afforess why has this been pulled if it removes existing functionality and does not offer a solution?

Afforess, line 103: Didn't notice. Use of /\* */ style comments means that the diff was not obvious. Sneaky.

Afforess, line 103: Also I suspect @kitskub is actually wrong. If I understand how ComponentHolder.get(...) works, you can request interfaces or superclasses and will have it return them if one exists.

windy1, line 103: That was my understanding as well.

kitskub, line 103: @Afforess no, you cannot do get(class<T) where T is not a component.
I disabled this for two reasons. First is the above reason. Second, NOTHING implements Container (if I am wrong, correct me).
Completely rewriting the get() method was out of the scope of my PR. I do, however, intend to make a PR that' allows the getting of interfaces.

windy1, line 103: Container has 12 implementing classes...

kitskub, line 103: Ok, guess my ide was being stupid. But still, still needs the other change.

windy1, line 103: This is an easy fix then. We just have to add a get method that takes a wildcard class reference instead of a type reference and also allow

https://github.com/SpoutDev/SpoutAPI/blob/master/src/main/java/org/spout/api/component/BaseComponentHolder.java#L213 

to take a wildcard reference. Of course, the trade of is that now you will have to cast your component.

kitskub, line 103: Yes.

Sleaker, line 103: just fyi, the redstonecomparator code was written with Viewers as a solution. I don't see why using viewers to determine block updates wouldn't work regardless of what you're doing with components.

kitskub, line 103: https://github.com/Vanilla/Vanilla/pull/611

PR #613, opened 2013-05-20
Fix WindowClickMessage and update for almost all inventory features for 1.5 changes 

kitskub:
The only feature not in this is number 5 in this video: http://www.youtube.com/watch?v=iZD5pJ2uIVM

### DIFF
--- a/src/main/java/org/spout/vanilla/inventory/window/AbstractWindow.java
+++ b/src/main/java/org/spout/vanilla/inventory/window/AbstractWindow.java
@@ -173,23 +173,22 @@ public abstract class AbstractWindow implements InventoryViewer {
 	/**
 	 * Gets the inventory at the specified native slot. Returns -1 if
 	 * non-existent.
-	 * @param nativeSlot clicked
+	 * @param nativeSlot clicked slot
 	 * @return inventory entry at slot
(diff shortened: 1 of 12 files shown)


Zidane:
Great commit but I have one little quirk with it.

Can you not use bytes (like for getClickArguments) and instead use enums? From API perspective, bytes are rather unfriendly...

Also could you catch up the javadocs entirely for windows while you are at it? Thanks :).

kitskub:
Ok, I'll do that.

kitskub:
I updated for the enum. Not sure what javadocs in particular you're thinking about. They seem pretty well documented.

windy1:
I believe he is asking you to document the enum. Conventional way to do this is.

``` java
/**
 * Description of foo
 **/
enum Foo {
    /**
     * Description
     **/
    BAR;
    /**
     *
     **/
    BAZ;

    // then document as you would a class
}
```

e: when I was quickly reading @Zidane 's comment I didn't see he was referring to Windows. Documentation for the enum would still be nice though.

Zidane:
I was talking about the javadocs for the Windowing System in Vanilla (in-general).

That can be another commit, thanks!

PR #618, opened 2013-05-29
Fixed the Camera

Waterpicker:
-Changed VanillaCamera into PlayerHead
-Camera uses Player Height and Head Rotation now
-Head.java's getRotation() was changed to getHeadRotation due to issues

Signed-off-by: Waterpicker Waterpicker@rocketmail.com


Zidane:
Pet peeve: Rename Head to EntityHead for me. Also detach those components FIRST then attach yours.

PR #623, opened 2013-06-02
Use inputState.pitch() instead of ts.getRotation().getPitch()

Grinch:
- Fixes up and down not being recognized but still reacts oddly when
  clicking on blocks with Spout Client


thehutch:
You should also change .multiply(speed).multiply(dt) to .multiply(speed \* dt) to prevent extra object creation

PR #640, opened 2013-08-11
Replace old manual metadata system with new automated MetadataComponent

bergerkiller:
In addition, also added the metadata that was missing for some entities. All metadata is now automatically updated by comparing live and synched values. Some of the old entity protocols have been removed as they ended up empty. Minecart/Boat has been further implemented to properly shake/take damage.

One possible issue that others will need to look into: joining Notchian servers with Spout Client and Vanilla. I added warning message when it tried to set metadata parameters for entities that is invalid. If you encounter any of these messages, let me know or try to fix it yourself. It warns about invalid indices or data type mismatches. If a Metadata entry is missing, it will likely log a warning for this; I added this to ease debugging.

I tested my changes on Spout server with minecarts: block changing works and I could damage the Minecart to cause it to 'shake'. I expect all other entities (e.g. sheep wool color) to work as well, but had no time to test that out properly.

Signed-off-by: Irmo van den Berge bergerkiller@gmail.com


windy1:
This makes much more sense and makes metadata easier to keep track of. :+1:

PR #641, opened 2013-08-12
Fix blast resistance values for all blocks

501st-alpha1:
Fixes VANILLA-410
Some blocks didn't have any resistance set, most just had wrong value.

### DIFF
--- a/src/main/java/org/spout/vanilla/material/block/liquid/Lava.java
+++ b/src/main/java/org/spout/vanilla/material/block/liquid/Lava.java
@@ -47,6 +47,7 @@ public class Lava extends Liquid implements InitializableMaterial {
 	public Lava(String name, int id, boolean flowing) {
 		super(name, id, flowing, VanillaMaterialModels.LAVA);
 		this.setFlowDelay(1500);
+		this.setResistance(flowing ? 0.0F : 500.0F);
 		if (getEngine().getPlatform() == Platform.CLIENT) {
 			if (!getModel().getRenderMaterial().getRenderEffects().contains(VanillaEffects.LIQUID)) {
 				getModel().getRenderMaterial().addRenderEffect(VanillaEffects.LIQUID);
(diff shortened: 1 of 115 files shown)


src/main/java/org/spout/vanilla/material/block/liquid/Lava.java
```diff
@@ -47,6 +47,7 @@
 	public Lava(String name, int id, boolean flowing) {
 		super(name, id, flowing, VanillaMaterialModels.LAVA);
 		this.setFlowDelay(1500);
+		this.setResistance(flowing ? 0.0F : 500.0F);
```
olliestanley, line 50: Idk if it matters too much, but this should probably be this.setFlowDelay(1500).setResistance(flowing ? 0.0F : 500F);
(Same for Water.java)

kitskub, line 50: tbh, there are many of Vanilla's materials that are this way. But to be sure, @Zidane?

olliestanley, line 50: Meh, best to keep it from getting any worse I guess :P

dredhorse:
I like the two liner

also tests are not working for this

olliestanley:
It's designed for the setters to return the object so you can do that in the first place. It should at least be kept consistent one way or the other. But really, this can always be changed later.

windy1:
Looks good, all we need to pull this is a sign-off to acknowledge you agree to the Certificate of Origin found [here](https://wiki.spout.org/Pull_request_guidelines#Certificate_of_origin). That is also the location of the broken link you posted in your post on spout.org as well.

You can edit a commit description by using `git commit --amend` and then adding `Signed-off-by: <you@example.com>` to the bottom of your description. You can update the commit here with `git push --force origin blast_resistance`.

501st-alpha1:
Okay, thanks. I'll take care of that when I get off work this evening. I can't figure out why the tests are failing, as they passed on my system. Ideas?

windy1:
I'm not sure, but I think it's unrelated to this PR anyways.

src/main/java/org/spout/vanilla/material/block/liquid/Lava.java
501st-alpha1, line 50: Yeah, I saw it both ways, so I wasn't sure which was preferred.

kitskub:
Oops sorry. I meant to say that test failures ARE unrelated to this PR.

PR #643, opened 2013-08-30
Fix JavaDoc wording for Ping.java

olliestanley:
Couldn't stop myself when I saw it :P

### DIFF
--- a/src/main/java/org/spout/vanilla/component/entity/player/Ping.java
+++ b/src/main/java/org/spout/vanilla/component/entity/player/Ping.java
@@ -34,7 +34,7 @@ import org.spout.vanilla.component.entity.VanillaEntityComponent;
 import org.spout.vanilla.event.player.network.PingEvent;
 
 /**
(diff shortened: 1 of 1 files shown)


windy1:
Sorry, but please don't open PRs for JavaDoc only changes. Either include it in another PR or just point it out to a team member.

olliestanley:
If you're going to have silly rules like that, they should at least be on spout.in/prguide so people actually know.

lukespragg:
It's the first I've heard of it. It's not a rule that I've approved though.

Generally it's more of the point the size, not that it's a doc PR.

lukespragg:
To add to my previous reply, we do accept Javadoc pull requests, but I'd
recommend putting more effort into making it have more substance, else just
mention the issue to a staff member if it's minor.

olliestanley:
Fair enough, but the size point should still be added to the pr guide.

lukespragg:
The whole doc needs to be redone honestly. One of the things on my todo
lost.

PR #645, opened 2013-10-08
Leaves on trees decay when logs burned by fire. Completes Vanilla-399

Amperial:
Need to fix the formatting. Still getting used to IntelliJ IDEA (used eclipse before)

### DIFF
--- a/src/main/java/org/spout/vanilla/material/block/misc/Fire.java
+++ b/src/main/java/org/spout/vanilla/material/block/misc/Fire.java
@@ -182,8 +182,12 @@ public class Fire extends VanillaBlockMaterial implements DynamicMaterial {
 				// Destroy the old block
 				if (mat instanceof TntBlock) {
 					((TntBlock) mat).onIgnite(sBlock, toCause(b)); // Ignite TntBlock
+				} else if (mat instanceof VanillaBlockMaterial) {
+					VanillaBlockMaterial sVanillaBlock = (VanillaBlockMaterial) mat; // Destroy vanilla blocks without drops
+					sVanillaBlock.getDrops().clear();
+					sVanillaBlock.destroy(sBlock, toCause(b));
 				} else {
-					sBlock.setMaterial(VanillaMaterials.AIR); // prevent drops
+					sBlock.setMaterial(VanillaMaterials.AIR); // Set non vanilla blocks to air
 				}
 				// Put fire in it's place?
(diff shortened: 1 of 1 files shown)


Amperial:
Formatting fixed, changes tested and confirmed working.

src/main/java/org/spout/vanilla/material/block/misc/Fire.java
```diff
@@ -182,8 +182,12 @@ public void onDynamicUpdate(Block b, long updateTime, int data) {
 				// Destroy the old block
 				if (mat instanceof TntBlock) {
 					((TntBlock) mat).onIgnite(sBlock, toCause(b)); // Ignite TntBlock
+				} else if (sBlock instanceof VanillaBlockMaterial) {
```
kitskub: `sBlock` is a `Block` which will never be a `VanillaBlockMaterial`. Replace `sBlock` with `mat`.

src/main/java/org/spout/vanilla/material/block/misc/Fire.java
```diff
@@ -182,8 +182,12 @@ public void onDynamicUpdate(Block b, long updateTime, int data) {
...
+					VanillaBlockMaterial sVanillaBlock = (VanillaBlockMaterial) sBlock; // Destroy vanilla blocks without drops
```
kitskub: Same thing here `sBlock` -> `mat`

Amperial:
That means it was still setting the logs to air, and the leaves actually burning instead of decaying..
Odd that it appeared as if they were decaying, since some of the leaves never actually got set on fire :P

lukespragg:
All good on this @Sleaker / @Zidane?

kitskub:
@ampayne2 the fact that it was "confirmed working" when the code was wrong makes me suspicious as to if this is really the correct "fix" or if it was even broken to begin with.

Amperial:
@kitskub it was confirmed working as far as I could tell, which involved burning a forest of trees and crashing every time an item dropped..
It was definitely broken to begin with because the leaf decay task is scheduled on log destroy, which setMaterial doesn't trigger. If getDrops().clear() on the VanillaBlockMaterial of the log works as I believe it does, the fix should work fine although i'm not completely sure since I didn't write the code for drops.
Making physics and item dropping more stable is kinda high priority so it's possible to fully test changes; testing this is hard enough, but testing my work on getting both iron and wooden doors to drop is impossible because they don't attempt to drop without physics enabled (which freezes the server).

kitskub:
@Sleaker @Zidane someone review this please

PR #646, opened 2013-10-13
Carpets now extend GroundAttachable instead of Solid. Completes Vanilla-...

Amperial:
...422

Signed-off-by: Austin Payne payneaustin5@gmail.com


Sleaker:
If carpet is burnable it needs to be able to support Fire, not just other carpet, looks like this was changed in 1.7 also:

13w24b  Carpets can be placed on top of any block (this includes blocks such as water source blocks, saplings, torches, chests, other carpet, etc. ).
This was changed in 1.6 - making these groundattachable is incorrect, they are solids.

PR #648, opened 2013-10-15
Carpet now breaks if block below destroyed. Completes Vanilla-422

Amperial:
Signed-off-by: Austin Payne payneaustin5@gmail.com

Sorry about this extra PR, had to make a separate branch for it so I could create a PR for paintings as well.


kitskub:
Looks good. Thanks.

PR #649, opened 2013-10-15
PaintingItem now implements Placeable. Completes Vanilla-425

Amperial:
Kills two birds with one stone:
- Solves the error of placing a painting where no paintings can be placed
- Takes the painting from your inventory because of implementing Placeable

It also makes more sense for PaintingItem to implement Placeable anyway, rather than using onInteract.


Amperial:
Any feedback on this?

kitskub:
Um. @Sleaker or @Zidane should review this.

kitskub:
@ampayne2 Has this been tested?

Amperial:
Yeah, it fixes the two problems it's meant to fix perfectly, but there's still one more painting bug left which is right clicking a painting crashes the server. It was already a bug though so this isn't the cause of that.

